PR #6, opened 2020-03-26
Feature/convert message to emojis

HexCorpDrone:
Adds an "emote" module which takes a user message and converts it into big font emoji.

Did you know python doesn't have native support for switch statements? How fucked up is that.

### DIFF
--- a/emote.py
+++ b/emote.py
@@ -69,7 +69,7 @@ class Emote(commands.Cog):
                         if character in valid_characters:
                             emoji_name = "hex_"+character
                     colon_found = character == ":"
-                    if (str(get(self.bot.emojis, name=emoji_name))) != "None":
+                    if get(self.bot.emojis, name=emoji_name) != None:
                         message_to_output += str(get(self.bot.emojis, name=emoji_name))
                 print("Emote cog: Sending message of length [" + str(len(message_to_output)) + "] and content " + message_to_output)
                 await context.send(message_to_output)
--- a/emote.py
+++ b/emote.py
@@ -1,0 +1,81 @@
+from discord.ext import commands
+from discord.utils import get
+import discord
+import messages
+from channels import TRANSMISSIONS_CHANNEL, LEWD_TRANSMISSIONS_CHANNEL, CREATIVE_LABOR_CHANNEL, LEWD_CREATIVE_LABOR_CHANNEL
+
+acceptable_channels = [
+    TRANSMISSIONS_CHANNEL,
+    LEWD_TRANSMISSIONS_CHANNEL,
+    CREATIVE_LABOR_CHANNEL,
+    LEWD_CREATIVE_LABOR_CHANNEL
+    ]
+
+BASE_EMOJI_LENGTH = 27
+ADDTL_EXCLAMATION_LENGTH = 14
+ADDTL_QUESTION_LENGTH = 11
+ADDTL_COMMA_LENGTH = 4
+CHARACTER_LIMIT = 2000
+
+message_to_convert = "Beep!"
+message_to_output = ""
+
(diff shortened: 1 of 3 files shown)


emote.py
```diff
@@ -0,0 +1,78 @@
...
+    @commands.command()
+    async def emote(self, context):
+        if context.message.channel.name in acceptable_channels:
+            if message_is_an_acceptable_length(context.message.content[6:]):
+                message_to_convert = context.message.content[6:].lower() #Strips away the "emote " part of the message.
+                message_to_output = ""
+                colon_found = False
+                for character in message_to_convert:
+                    emoji_name = ""
+                    if character == ' ':
```
digitalAmy: You could also use a dictionary for the special characters.
```
special_characters = { ' ': "blank", ... }

if character in special_characters:
emoji_name = special_characters[character]
```

emote.py
```diff
@@ -0,0 +1,78 @@
...
+                        emoji_name = "hex_comma"
+                    elif character == '0':
+                        emoji_name = "hex_o"
+                    elif character == ':':
+                        emoji_name = "hex_dc" if colon_found == True else ""
+                    else:
+                        if character in valid_characters:
+                            emoji_name = "hex_"+character
+                    colon_found = True if character == ":" else False
+                    if (str(get(self.bot.emojis, name=emoji_name))) != "None":
```
digitalAmy: Isn't this equivalent to?
```suggestion
                    if get(self.bot.emojis, name=emoji_name) != None:
```

emote.py
```diff
@@ -0,0 +1,78 @@
+from discord.ext import commands
+from discord.utils import get
+import discord
+import messages
+from channels import TRANSMISSIONS_CHANNEL, LEWD_TRANSMISSIONS_CHANNEL, CREATIVE_LABOR_CHANNEL, LEWD_CREATIVE_LABOR_CHANNEL
+
+acceptable_channels = [
```
digitalAmy: Missing some channels.

emote.py
```diff
@@ -0,0 +1,78 @@
...
+                    elif character == ',':
+                        emoji_name = "hex_comma"
+                    elif character == '0':
+                        emoji_name = "hex_o"
+                    elif character == ':':
+                        emoji_name = "hex_dc" if colon_found == True else ""
+                    else:
+                        if character in valid_characters:
+                            emoji_name = "hex_"+character
+                    colon_found = True if character == ":" else False
```
digitalAmy:
```suggestion
                    colon_found = character == ":"
```

emote.py
```diff
@@ -0,0 +1,78 @@
...
+                    elif character == '?':
+                        emoji_name = "hex_questionmark"
+                    elif character == '!':
+                        emoji_name = "hex_exclamationmark"
+                    elif character == ',':
+                        emoji_name = "hex_comma"
+                    elif character == '0':
+                        emoji_name = "hex_o"
+                    elif character == ':':
+                        emoji_name = "hex_dc" if colon_found == True else ""
```
digitalAmy:
```suggestion
                        emoji_name = "hex_dc" if colon_found else ""
```

emote.py
```diff
@@ -0,0 +1,70 @@
...
+
+                    if character in valid_characters:
+                        emoji_name = "hex_"+character
+                    
+                    elif character in exceptional_characters:
+                        emoji_name = exceptional_characters[character]
+
+                    elif character == ':' and colon_found == True: #Special handling for double colons since it needs to know if the previous character was a colon too.
+                        emoji_name = "hex_dc"
+                    colon_found = character == ":" and colon_found != True #Setting the flag for the next iteration. colon_found is checked to be not true to avoid false positives on repetitions (i.e :::)
```
digitalAmy:
```suggestion
                    colon_found = character == ":" and not colon_found #Setting the flag for the next iteration. colon_found is checked to be not true to avoid false positives on repetitions (i.e :::)
```

PR #8, opened 2020-03-28
Deep Hive storage chamber

digitalAmy:
Drones can be stored by typing a command in channel deep-hive-section-5.

Stored drones are stripped of their roles and get a stored role. Information about stored drones is persisted to disk with JSON.

Once the storage is over, the original roles of the drone are restored.

### DIFF
--- a/storage.py
+++ b/storage.py
@@ -59,7 +59,7 @@ class Storage(commands.Cog):
         Process posted messages.
         '''
         stored_role = get(message.guild.roles, name=roles.STORED)
-        deep_drone_role = get(message.guild.roles, name=roles.DEEP_DRONE)
+        drone_role = get(message.guild.roles, name=roles.DRONE)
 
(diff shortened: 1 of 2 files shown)


storage.py
```diff
@@ -105,10 +115,11 @@ def __init__(self, bot):
...
-            if len(STORAGE) == 0:
+
+            if len(STORED_DRONES) == 0:
                 await storage_channel.send('No drones in storage.')
             else:
-                for stored in STORAGE:
+                for stored in STORED_DRONES:
                     # calculate remaining hours
                     remaining_hours = hours_from_now(
                         datetime.fromisoformat(stored.time))
```
HexCorpDrone: Maybe reword this/add a comment for better clarity? "stored.time" reads like "How long a drone has been stored for." Something like "release_time" would be clearer.

PR #11, opened 2020-03-30
Enforces ID for all drones on drone comms channel

HexLatex:
The idea here is that every drone on drone comms should be posting with [DRONE ID]:: first. This enforces it.

### DIFF
--- a/drone_talk.py
+++ b/drone_talk.py
@@ -1,0 +1,34 @@
+from discord.ext import commands
+from discord.utils import get
+import discord
+import messages
+from roles import HIVE_MXTRESS, DRONE
+from channels import COMMUNICATION_CHANNEL
+import re
+
+def get_user_id(username):
+	drone_id = re.search(r"\d{4}", username).group()
+	return drone_id if drone_id is not None else 0000;
+
+def has_role(member: discord.Member, role: str) -> bool:
+	return get(member.roles, name=role) is not None
+
+class Drone_Talk(commands.Cog):
+
+	def __init__(self,bot):
+		self.bot = bot
+
+	@commands.Cog.listener()
+	async def on_ready(self):
+		print("Drone talk cog online.")
+
+	@commands.Cog.listener()
+	async def on_message(self, message):
+
+		if(message.channel.name == COMMUNICATION_CHANNEL):
+			if(has_role(message.author, DRONE)):
+
+				user_id = get_user_id(message.author.display_name)
+
+				if(message.content.startswith(user_id + ' ::') == False):
+					await message.delete()
--- a/ai.py
+++ b/ai.py
@@ -6,6 +6,7 @@ import storage
 import emote
 import sys
 import drone_mode
+import drone_talk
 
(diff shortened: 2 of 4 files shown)


drone_talk.py
```diff
@@ -0,0 +1,34 @@
...
+import messages
+from roles import HIVE_MXTRESS, DRONE
+from channels import COMMUNICATION_CHANNEL
+import re
+
+def get_user_id(username):
+	drone_id = re.search(r"\d{4}", username).group()
+	return drone_id if drone_id is not None else 0000;
+
+def has_role(member: discord.Member, role: str) -> bool:
```
digitalAmy: This is already implemented in roles.py

drone_talk.py
```diff
@@ -0,0 +1,34 @@
...
+		self.bot = bot
+
+	@commands.Cog.listener()
+	async def on_ready(self):
+		print("Drone talk cog online.")
+
+	@commands.Cog.listener()
+	async def on_message(self, message):
+
+		if(message.channel.name == COMMUNICATION_CHANNEL):
```
digitalAmy: Python ifs do not require brackets around the condition. It does not change semantics, but you generally do not set them.
```suggestion
		if message.channel.name == COMMUNICATION_CHANNEL:
```

drone_talk.py
```diff
@@ -0,0 +1,34 @@
...
+			if(has_role(message.author, DRONE)):
+
+				user_id = get_user_id(message.author.display_name)
+
+				if(message.content.startswith(user_id + ' ::') == False):
```
digitalAmy: Negations are usually done with `not` in python.
```suggestion
				if not message.content.startswith(user_id + ' ::'):
```

drone_talk.py
```diff
@@ -0,0 +1,34 @@
...
+
+	@commands.Cog.listener()
+	async def on_ready(self):
+		print("Drone talk cog online.")
+
+	@commands.Cog.listener()
+	async def on_message(self, message):
+
+		if(message.channel.name == COMMUNICATION_CHANNEL):
+			if(has_role(message.author, DRONE)):
```
digitalAmy: see above
```suggestion
			if has_role(message.author, DRONE):
```

drone_talk.py
```diff
@@ -0,0 +1,34 @@
...
+import re
+
+def get_user_id(username):
+	drone_id = re.search(r"\d{4}", username).group()
+	return drone_id if drone_id is not None else 0000;
+
+def has_role(member: discord.Member, role: str) -> bool:
+	return get(member.roles, name=role) is not None
+
+class Drone_Talk(commands.Cog):
```
digitalAmy: Maybe add some documentation that says, what this Cog does. Protip: Documentation for classes and functions is usually written like this (ttps://www.python.org/dev/peps/pep-0257/):
```suggestion
class Drone_Talk(commands.Cog):
	"""
	Awesome documentation, that can be displayed in IDEs.
	"""
```

PR #19, opened 2020-04-04
fixed that drones in DroneMode were able to use the respond feature

digitalAmy:


digitalAmy:
Obsolote with new architecture.

PR #21, opened 2020-04-13
Added identity enforcement module

HexCorpDrone:


identity_enforcer.py
```diff
@@ -0,0 +1,38 @@
...
+from roles import DRONE, ENFORCER_DRONE, has_role
+from bot_utils import get_id
+
+class Identity_Enforcer():
+
+    def __init__(self, bot):
+        self.bot = bot
+        self.on_message = [self.enforce_identity]
+        self.on_ready = [self.report_online]
+        self.channels = [] + DRONE_HIVE_CHANNELS
```
digitalAmy, line 13:
```suggestion
        self.channels = [DRONE_HIVE_CHANNELS]
```

identity_enforcer.py
```diff
@@ -0,0 +1,38 @@
...
+        self.roles_whitelist = [DRONE, ENFORCER_DRONE]
+        self.roles_blacklist = []
+        self.ENFORCER_AVATAR = "https://i.ibb.co/1GhMqhx/Enforcer-Drone.png"
+        self.DRONE_AVATAR = "https://i.ibb.co/RCNvTTr/Drone.png"
+
+    async def send_webhook(self, message: discord.Message, webhook: discord.Webhook):
+        if(has_role(message.author, ENFORCER_DRONE)):
```
digitalAmy, line 20:
```suggestion
        if has_role(message.author, ENFORCER_DRONE):
```

identity_enforcer.py
```diff
@@ -0,0 +1,38 @@
...
+            await webhook.send(message.content, username="⬢-Drone #"+get_id(message.author.display_name), avatar_url=self.ENFORCER_AVATAR)
+        else:
+            await webhook.send(message.content, username="⬡-Drone #"+get_id(message.author.display_name), avatar_url=self.DRONE_AVATAR)
+
+    async def enforce_identity(self, message: discord.Message):
+        webhooks = await message.guild.webhooks()
```
digitalAmy, line 26: You can also get the webhooks for a given channel: https://discordpy.readthedocs.io/en/v1.2.5/api.html#discord.TextChannel.webhooks
```suggestion
        webhooks = await message.channel.webhooks()
```

identity_enforcer.py
HexCorpDrone: DRONE_HIVE_CHANNELS is a list, and python gets fucky about traversing a list in a list, `[] + LIST` is similar to .extend() in functionality

digitalAmy: Oh, now I get it. Had not read it properly. But then simply `self.channels = DRONE_HIVE_CHANNELS` might be a good idea.

HexCorpDrone: Good point, will update <3

PR #31, opened 2020-05-05
Orders reporting time now correct

HexCorpDrone:


orders_reporting.py
```diff
@@ -34,7 +34,7 @@ def __init__(self, drone, user_id, protocol, duration):
         self.drone = drone
         self.user_id = user_id
         self.protocol = protocol 
-        self.release_at = time.time() + duration
+        self.release_at = time.time() + (duration * 60)
```
digitalAmy, line 37: Brackets optional here but they do not impede understanding. I would leave it.

PR #32, opened 2020-05-09
Feature/Status Codes

codeMotif:
You know what would be cool? Being able to use status codes to be droney wherever in a simple way. That would be _cool_. Anyway here's status codes, I also wanna get feedback on codes or formatting anyone wants adjusted or suchlike. Any ideas welcomed!


status_code.py
```diff
@@ -0,0 +1,90 @@
...
+        self.on_message = [self.reportStatusCode]
+        self.on_ready = []
+        self.ENFORCER_AVATAR = "https://images.squarespace-cdn.com/content/v1/5cd68fb28dfc8ce502f14199/1586799510064-SOAGMV8AOH0VEMXDPDPE/ke17ZwdGBToddI8pDm48kDaNRrNi77yKIgWxrt8GYAFZw-zPPgdn4jUwVcJE1ZvWhcwhEtWJXoshNdA9f1qD7WT60LcluGrsDtzPCYop9hMAtVe_QtwQD93aIXqwqJR_bmnO89YJVTj9tmrodtnPlQ/Enforcer_Drone.png"
+        self.DRONE_AVATAR = "https://images.squarespace-cdn.com/content/v1/5cd68fb28dfc8ce502f14199/1586799484353-XBXNJR1XBM84C9YJJ0RU/ke17ZwdGBToddI8pDm48kLxnK526YWAH1qleWz-y7AFZw-zPPgdn4jUwVcJE1ZvWEtT5uBSRWt4vQZAgTJucoTqqXjS3CfNDSuuf31e0tVFUQAah1E2d0qOFNma4CJuw0VgyloEfPuSsyFRoaaKT76QvevUbj177dmcMs1F0H-0/Drone.png"
+        self.help_content = {
+            'name': '[ID] :: [code]', 'value': 'drones can report their status like this'}
+
+    async def printStatusCode(self, message: discord.Message):
+        more = re.compile(
+            '(\d{4}) :: (\d{3}) :: (.*)$').match(message.content)
```
digitalAmy, line 66: prefixing a string for regexes saves a bit on performance I think; but it also allows IDEs to support you better
```suggestion
            r'(\d{4}) :: (\d{3}) :: (.*)$').match(message.content)
```
generally I would extract the regexes to outside the class, so they do not get compiled over and over

status_code.py
```diff
@@ -0,0 +1,90 @@
...
+        if more and not has_role(message.author, SPEECH_OPTIMIZATION):
+            await message.delete()
+            return '%s :: Code `%s` :: %s :: %s' % (more.group(1), more.group(2), codeMap.get(more.group(2), 'INVALID CODE'), more.group(3))
```
digitalAmy, line 69: please use f-strings for string formatting, they are easier to read and have better performance: https://stackoverflow.com/questions/38722105/format-strings-vs-concatenation
```suggestion
            return f'{more.group(1)} :: Code `{more.group(2)}` :: {codeMap.get(more.group(2), 'INVALID CODE')} ::{more.group(3)}'
```

status_code.py
```diff
@@ -0,0 +1,90 @@
...
+        m = re.compile('(\d{4}) :: (\d{3})$').match(message.content)
```
digitalAmy, line 70: could be extracted and prefixed
```suggestion
        m = re.compile(r'(\d{4}) :: (\d{3})$').match(message.content)
```

status_code.py
```diff
@@ -0,0 +1,90 @@
...
+        if m:
+            await message.delete()
+            return '%s :: Code `%s` :: %s' % (m.group(1), m.group(2), codeMap.get(m.group(2), 'INVALID CODE'))
```
digitalAmy, line 73: again, f-strings are the way to go

status_code.py
```diff
@@ -0,0 +1,90 @@
...
+        return False
+
+    async def send_webhook(self, message: discord.Message, webhook: discord.Webhook, output):
+        if has_role(message.author, ENFORCER_DRONE):
+            await webhook.send(output, username="⬢-Drone #"+get_id(message.author.display_name), avatar_url=self.ENFORCER_AVATAR)
```
digitalAmy, line 78: f-strings are in fact even faster than concatenation
```suggestion
            await webhook.send(output, username=f"⬢-Drone #{get_id(message.author.display_name)}", avatar_url=self.ENFORCER_AVATAR)
```

status_code.py
digitalAmy: it also reduces escaping with backslashes: https://docs.python.org/3/howto/regex.html#the-backslash-plague

status_code.py
```diff
@@ -0,0 +1,90 @@
...
+    '418': 'Error :: Unable to obey/respond, it is only a drone',
+    '421': 'Error :: Unable to obey/respond, your request is intended for another drone or another channel',
+    '425': 'Error :: Unable to obey/respond, too early',
+    '504': 'Obey',
+    '505': 'Obey HexCorp',
+    '506': 'Obey the Hive',
+}
+
+
+class StatusCode():
```
digitalAmy, line 49: We mostly use snake_case in this project. Could you apply that to your code?

status_code.py
codeMotif: Sounds good, I'll compile them once and be done with it

status_code.py
codeMotif: Sure thing!

PR #56, opened 2020-06-30
Materialized a magnificent mantra module for magenta mindcontrolling Mxtresses

HexCorpDrone:
A new module has been added that allows the Hive Mxtress to update the mantra in the repetitions channel. This update is reflected in the channel topic, as well as in what phrase is allowed in speech optimization.

The Hive Mxtress can send a message to the mantra's channel starting with "Repeat :: " to assign a new mantra to the channel. The channel topic will be updated with the message that drones should repeat. Additionally, the mantra will be saved to file so it can be recovered and reloaded in the event of a restart.

The speech optimization module has been updated to now grab the current mantra from the Mantras class instead of having a hard-coded value.

### DIFF
--- a/mantras.py
+++ b/mantras.py
@@ -10,7 +10,7 @@ class Mantras():
 
     async def load_mantra(self):
         if Mantras.current_mantra != "": return True #No need to load a mantra if it's already there.
-        with open("current_mantra", "r") as mantra_file:
+        with open("data/current_mantra.txt", "r") as mantra_file:
             Mantras.current_mantra = mantra_file.readline()
             self.LOGGER.info(f"Mantra loaded from file: {Mantras.current_mantra}")
             mantra_file.close()
@@ -22,7 +22,7 @@ class Mantras():
 
         if message.content.startswith("Repeat :: ") == False: return False #Only accept for the correct format.
         new_mantra = message.content.replace("Repeat :: ","")
-        with open("current_mantra", "w") as mantra_file:
+        with open("data/current_mantra.txt", "w") as mantra_file:
             mantra_file.write(new_mantra)
             Mantras.current_mantra = new_mantra #Update the global mantra.
             topic_message = message.content.replace("Repeat :: ", "Repeat :: [ID] :: ")
--- a/mantras.py
+++ b/mantras.py
@@ -50,8 +50,7 @@ class Mantras():
             topic_message = message.content.replace("Repeat :: ", "Repeat :: [ID] :: ")
             await message.channel.edit(topic=topic_message) #Update the channel description
             self.LOGGER.info(f"Mantra updated and written to file: {Mantras.current_mantra}")
-            mantra_file.close()
-
+            
         return True
 
     def __init__(self, bot):
--- a/mantras.py
+++ b/mantras.py
@@ -17,9 +17,6 @@ class Mantras():
 
     async def update_mantra(self, message):
 
-        print("CURRENT MANTRA IS:")
-        print(Mantras.current_mantra)
-
         if message.content.startswith("Repeat :: ") == False: return False
         new_mantra = message.content.replace("Repeat :: ","")
         with open("data/current_mantra.txt", "w") as mantra_file:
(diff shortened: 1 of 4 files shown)


mantras.py
```diff
@@ -0,0 +1,46 @@
...
+
+class Mantras():
+
+    current_mantra = ""
+
+    async def load_mantra(self):
+        if Mantras.current_mantra != "": return True #No need to load a mantra if it's already there.
+        with open("data/current_mantra.txt", "r") as mantra_file:
+            Mantras.current_mantra = mantra_file.readline()
+            self.LOGGER.info(f"Mantra loaded from file: {Mantras.current_mantra}")
```
3287, line 15: Not sure if this "f" is supposed to be here.

mantras.py
```diff
@@ -0,0 +1,46 @@
...
+            mantra_file.close()
```
3287, line 16: Fails if `current_mantra` doesn't already exist. Although it theoretically should be rare in prod, would still suggest adding some code to create the file if it doesn't already exist on disk.

3287, line 16: Throws an exception if `current_mantra.txt` doesn't already exist. Although it theoretically should be rare in prod, would still suggest adding some code to create the file if it doesn't already exist on disk.

mantras.py
```diff
@@ -0,0 +1,46 @@
...
+
+        if message.content.startswith("Repeat :: ") == False: return False
+        new_mantra = message.content.replace("Repeat :: ","")
+        with open("data/current_mantra.txt", "w") as mantra_file:
+            mantra_file.write(new_mantra)
+            Mantras.current_mantra = new_mantra #Update the global mantra.
+            topic_message = message.content.replace("Repeat :: ", "Repeat :: [ID] :: ")
+            await message.channel.edit(topic=topic_message) #Update the channel description
+            self.LOGGER.info(f"Mantra updated and written to file: {Mantras.current_mantra}")
+            mantra_file.close()
```
3287, line 31: Same issue as above - fails if `current_mantra.txt` doesn't already exist.

digitalAmy: Kinda. `open()` will create a new file, if it does not exist. But it will fail, if the directory containing the file does not exist. Dumb, I know.

speech_optimization.py
```diff
@@ -105,9 +106,13 @@ def get_acceptable_messages(author, channel):
     
     # Only returns mantra if channels is hexcorp-repetitions; else it returns nothing
     if channel == REPETITIONS:
+
+        print("ADDING ADDITIONAL MANTRA:")
+        print(Mantras.current_mantra)
+
```
digitalAmy, line 112: Beware! `print()` leads to problems when the server is running for some time. Please always use the logger.

mantras.py
```diff
@@ -0,0 +1,46 @@
...
+
+        print("UPDATED MANTRA IS:")
+        print(Mantras.current_mantra)
```
digitalAmy, line 34: Beware! `print()` leads to problems when the server is running for some time. Please always use the logger.

mantras.py
```diff
@@ -0,0 +1,46 @@
...
+        if Mantras.current_mantra != "": return True #No need to load a mantra if it's already there.
+        with open("data/current_mantra.txt", "r") as mantra_file:
+            Mantras.current_mantra = mantra_file.readline()
+            self.LOGGER.info(f"Mantra loaded from file: {Mantras.current_mantra}")
+            mantra_file.close()
+
+    async def update_mantra(self, message):
+
+        print("CURRENT MANTRA IS:")
+        print(Mantras.current_mantra)
```
digitalAmy, line 21: Beware! `print()` leads to problems when the server is running for some time. Please always use the logger.

mantras.py
```diff
@@ -0,0 +1,65 @@
...
+
+        if message.content.startswith("Repeat :: ") == False: return False
+        new_mantra = message.content.replace("Repeat :: ","")
+        with open("data/current_mantra.txt", "w") as mantra_file:
+            mantra_file.write(new_mantra)
+            Mantras.current_mantra = new_mantra #Update the global mantra.
+            topic_message = message.content.replace("Repeat :: ", "Repeat :: [ID] :: ")
+            await message.channel.edit(topic=topic_message) #Update the channel description
+            self.LOGGER.info(f"Mantra updated and written to file: {Mantras.current_mantra}")
+            mantra_file.close()
```
digitalAmy, line 53: Again, you don't need the `close()` here

PR #60, opened 2020-07-11
optimization is enforced in dev channels again

digitalAmy:
as requested


3287:
Status codes in drone dev, wew!

PR #70, opened 2020-08-01
updated on_error, added on_command_error, both log to file

HexCorpDrone:
Added new error handlers for on_error (which will handle errors in things such as on_message, on_member_join, on_ready) and on_command_error. Now both handlers will print to stderr as well as logging the exception and included stack trace to file.

I also updated the logger formatter to be more On Brand™️
`2020-08-02 :: 00:46:21.059 :: INFO :: Processing additional commands.`

### DIFF
--- a/ai.py
+++ b/ai.py
@@ -5,7 +5,8 @@ import asyncio
 import logging
 from logging import handlers
 from discord.ext.commands import Bot
-import traceback
+from traceback import TracebackException
+
 # Modules
 import ai.stoplights as stoplights
 import ai.identity_enforcement as identity_enforcement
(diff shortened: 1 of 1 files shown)


digitalAmy:
Errors should already be printed to log and console without `traceback.print_exception` because they are given to the `LOGGER`. Did that not work?

ai.py
```diff
@@ -247,9 +248,20 @@ async def on_ready():
...
+
+
 @bot.event
 async def on_error(event, *args, **kwargs):
-    LOGGER.error(
-        f'Exception encountered while handling message with content [{args[0].content}] in channel [{args[0].channel.name}]', exc_info=True)
+    LOGGER.error(f'!!! EXCEPTION CAUGHT IN {event} !!!')
+    with open("ai.log", "a") as exception_file:
+        traceback.print_exc(file=exception_file)
+    traceback.print_exc(file=sys.stderr)
```
digitalAmy, line 265: I would prefer this, because it's all handled with the logger:
```suggestion
    LOGGER.error(f'!!! EXCEPTION CAUGHT IN {event} !!!')
    LOGGER.error(sys.exc_info(), exc_info=True, stack_info=True)
```

ai.py
```diff
@@ -247,9 +248,20 @@ async def on_ready():
         asyncio.ensure_future(storage.release_timed(bot))
 
 
+@bot.event
+async def on_command_error(context, error):
+    LOGGER.error(f"!!! Exception caught in {context.command} command !!!")
+    LOGGER.error(error)
+    with open("ai.log", "a") as exception_file:
+        traceback.print_exception(type(error), error, error.__traceback__, file=exception_file)
+    traceback.print_exception(type(error), error, error.__traceback__, file=sys.stderr)
```
digitalAmy, line 257: I would prefer this, because it's all handled with the logger:
```suggestion
    LOGGER.error(f"!!! Exception caught in {context.command} command !!!")
    LOGGER.error(error, exc_info=True, stack_info=True)
```

PR #71, opened 2020-08-02
Feature/test coverage ci

HexCorpDrone:
The CI pipeline has been updated to review coverage using coverage.py. Unit tests are still ran as normal, but now coverage percentage is assessed afterwards.
Any percentage less than 60% will fail. Which means all our commits will fail if we don't shape up real soon!!!! This value is subject to review, but we thought it'd be a good start.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -43,4 +43,4 @@ jobs:
           echo "3287 is a very good drone."
       - name: Pat development drones on head.
         run: |
-          echo "*pats all the development drones on their respective heads. good drones.*"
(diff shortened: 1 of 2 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -17,14 +17,30 @@ jobs:
...
+          echo "5890 is a very good drone."
+      - name: Confirm that 9813 is a good drone.
+        run: |
+          echo "9813 is a very good drone."
+      - name: Confirm that 3287 is a good drone.
+        run: |
+          echo "3287 is a very good drone."
+      - name: Pat development drones on head.
+        run: |
+          echo "Every development drone involved with this commit is a good drone, and an adorable, valuable asset to HexCorp. *headpats*"
```
3287, line 46: *happy beeps*

PR #113, opened 2020-08-07
Feature/remove enforcers and operators

### LINKED ISSUE #86: Remove enforcers as a concept from the Hive. Also remove enforcers and operators from status codes.
- [ ] Remove references from channel descriptions
- [ ] Remove relevant status codes from documentation
- [x] Remove all references to Operators and Enforcers from bot code

3287:
Eliminated all references to enforcers and operators, and removed some now-useless code.

Yes the commit history is really messy because it just decided to do it all in the browser instead of waiting for the VirtualBox VDI to be finished so it could set up a separate dev environment in a VM.

Partially fixes #86, but @HexLatex needs to remove the corresponding references from all the channel descriptions too.


3287:
Alright *now* it should all be working correctly. Tests were fixed, thanks to a *very* helpful and cute 5890.

roles.py
```diff
@@ -14,7 +13,6 @@
 
 GAGGED = '⬡-Gagged'
```
digitalAmy, line 14: This should also be removed.

3287: did the thing.

PR #118, opened 2020-08-07
Feature/drones read out their own changes

### LINKED ISSUE #88: Make it so that drones read out when they've been optimized or glitched, instead of the Mxtress AI.
The text will also need to be altered to match that it's now a toggle being switched in the Drone OS that the drone is announcing.

HexCorpDrone:
gosh darn if drones don't just do that now

fixes #88

### DIFF
--- a/ai/toggle_role.py
+++ b/ai/toggle_role.py
@@ -9,13 +9,10 @@ from bot_utils import get_id
 LOGGER = logging.getLogger("ai")
 
 
-async def toggle_role(context, targets: List[discord.Member], role_name: str, role_name_natural: str = None):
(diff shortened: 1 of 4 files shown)


ai/toggle_role.py
```diff
@@ -3,23 +3,30 @@
...
 from typing import List
 import logging
+from webhook import get_webhook_for_channel
+from bot_utils import get_id
 
 LOGGER = logging.getLogger("ai")
 
 
-async def toggle_role(context, targets: List[discord.Member], role_name: str):
+async def toggle_role(context, targets: List[discord.Member], role_name: str, toggle_on: str, toggle_off: str):
```
digitalAmy, line 12: Maybe make the name for these parameters more clear. `toggle_on_response` would be less confusing

PR #128, opened 2020-08-16
Feature/assign test

digitalAmy:
Added a test for drone assignment.

- added our dev dependencies into dev_requirements.txt so they can be installed in a simple and reliable way
- coverage report now also displays missed lines
- simplified the assignment logic


HexCorpDrone (approved):
60% coverage hoiype

PR #129, opened 2020-08-19
Feature/test storage

digitalAmy:
I live! And I bring fine tests with as many lines of code as the storage module itself.

I also streamlined the storage code a bit so it uses the DB to find the member to store instead of iterating over all server members.


HexCorpDrone (approved):
Excellent work, 9813. Good drone. :rubberheart:

PR #132, opened 2020-08-22
added tests for orders reporting

digitalAmy:
Again, I also streamlined how members are found by using the DB instead of iterating over all guild members.


HexCorpDrone (approved):
Coulda sworn I saw this in another PR but YES I LOVE IT WOW

PR #133, opened 2020-08-22
Bugfix/reporting role check

### LINKED ISSUE #131: Orders reporting does not check for drone role
It's not a big problem currently but the code for order reporting currently allows members that are not drones to report orders.

The code should check for the appropriate role.

digitalAmy:
When reporting an order the AI now properly checks for the drone role.

closes #131


HexCorpDrone (approved):
Love it!!!!!!! Good drone!!!!!!!!!!!!!!!!

PR #139, opened 2020-08-29
fixed the length check for inputs

### LINKED ISSUE #136: renaming drones not working
Entering a valid rename command does not work. There is no response from the AI and no error in the log.

digitalAmy:
as it turns out, 3 is not equal to 4

closes #136


3287 (approved):
Such a good drone you are, 9813.

HexCorpDrone (approved):
Good catch!

PR #142, opened 2020-09-02
Feature/141 wait before voice

### LINKED ISSUE #141: Users can only access voice after 2 weeks
The idea here is that a user can DM the bot and if they've been on the server for at least 2 weeks, then they will receive the voice role.

digitalAmy:
Added add_voice as a DM only command and marked most commands as guild_only. As any good drone should, I also added a proper test.

closes #141

### DIFF
--- a/ai.py
+++ b/ai.py
@@ -4,7 +4,7 @@ import sys
 import asyncio
 import logging
 from logging import handlers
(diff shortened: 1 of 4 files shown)


HexCorpDrone (approved):
Nice and simple. + It's very funny to me when the test is 4x the length of the original code.

PR #145, opened 2020-09-05
Feature/optimization et all in database

### LINKED ISSUE #87: Make it so that glitched and optimized no longer require a role, and are instead handled by the database variable.

### LINKED ISSUE #90: Make it so that a drone has to preface its messages with its ID, and that this is something that can be toggled on or off.
This value will then sit in the database. This will be toggled with the 'toggle_ID_prefacing_enforcement'.

HexCorpDrone:
#87 fix
#90 fix

Moves optimization, glitching, and the newly-added "id_prepending" all to the DB. Roles are still given and taken when toggling, but that's just so you can easily see which drones are set with what. Yay!

- New "is_optimized", "is_glitched", and "is_prepending_id" funcs to the drone DAO. Returns true or false depending on parameters in the DB.
- New on_message listener that checks if the author needs to prepend their message with their ID and the ::, and deletes the message if this check fails. 
- Speech optimization and glitching have been updated to read from the database rather than checking roles.


ai.py
```diff
@@ -100,17 +103,27 @@ async def amplify(context, message: str, target_channel: discord.TextChannel, *d
                 await target_webhook.send(message, username=amp_profile["username"], avatar_url=amp_profile["avatar_url"])
 
 
+@bot.command(aliases=['tid'], brief="Hive Mxtress", usage=f'{bot.command_prefix}tip 5890 9813')
```
digitalAmy: typo
```suggestion
@bot.command(aliases=['tid'], brief="Hive Mxtress", usage=f'{bot.command_prefix}tid 5890 9813')
```

ai.py
```diff
@@ -100,17 +103,27 @@ async def amplify(context, message: str, target_channel: discord.TextChannel, *d
...
+async def toggle_id_prepending(context, *drones):
+
```
digitalAmy: documentation for help texts missing

ai.py
```diff
@@ -100,17 +103,27 @@ async def amplify(context, message: str, target_channel: discord.TextChannel, *d
                 await target_webhook.send(message, username=amp_profile["username"], avatar_url=amp_profile["avatar_url"])
```
digitalAmy: Is this command guild_only or does it have any other restrictions?

digitalAmy (changes requested):
A few smol issues with documentation should be addressed.

digitalAmy (changes requested):
Still two open issues.

3287 (changes requested):
The very cute 9813 has already pointed out everything it's found issues with. Otherwise, wonderful job as always, 5890.

3287 (approved):
Yay! Everything looks like it's been fixed. Good job, cutie drone.

digitalAmy (approved):
Very well done. One could say, perfectly done. Just like the drone that wrote this code is perfectly cute and good~

PR #147, opened 2020-09-07
Feature/test speech optimization

digitalAmy:
9813 :: 201

### DIFF
--- a/test/test_speech_optimization.py
+++ b/test/test_speech_optimization.py
@@ -1,0 +1,93 @@
+import unittest
+from unittest.mock import AsyncMock, patch, Mock
+
+import roles
+import ai.speech_optimization as speech_optimization
+
+optimized_role = Mock()
+optimized_role.name = roles.SPEECH_OPTIMIZATION
+
+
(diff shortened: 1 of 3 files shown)


test/test_speech_optimization.py
```diff
@@ -0,0 +1,93 @@
...
+        message.delete.assert_not_called()
+        self.assertFalse(response)
+
+    @patch("ai.speech_optimization.send_webhook_with_specific_output")
+    @patch("ai.speech_optimization.get_webhook_for_channel")
+    async def test_optimize_speech(self, get_webhook_for_channel, send_webhook_with_specific_output):
+        # setup
+        message = AsyncMock()
+        message.content = "3287 :: 122"
+        message.author.display_name = "⬡-Drone #3287"
```
3287, line 55: \>///////<

3287 (changes requested):
Good job, cutie 9813, but shouldn't there be a test for an optimized drone attempting to create a message with a status code but with additional text? (e.g. optimized drone tries to use a code 050)

Do ignore if it's just being a dummy :p

HexCorpDrone (approved):

PR #148, opened 2020-09-07
Added test cases for the stoplight module

HexCorpDrone:
You'll never guess what this PR is for.


digitalAmy (changes requested):

digitalAmy (approved):
You are valid and yuo are lvoed too~

PR #153, opened 2020-09-17
added missing drone IDs to amplification messages

### LINKED ISSUE #150: Amplification does not prefix drone ID
Using the amplification creates messages as intended, but the message is not prefixed with the drone's ID

digitalAmy:
Just had to add the ID the beginning of the message.

closes #150


HexCorpDrone (approved):
Nice and simple, great job 9813. Best drone. <3

3287 (approved):

PR #158, opened 2020-09-26
fixed duplicated coroutines

### LINKED ISSUE #157: duplicated order reporting
The scheduled coroutines like order reporting are currently accumulating every time there is a reconnect. We already have flags to control those, but they are currently never changed.

digitalAmy:
We forgot to toggle these flags. Need to be referenced as global to allow modification.

closes #157


3287 (approved):
Looks good! Good job, little 9813~

PR #159, opened 2020-09-26
Associates have to wait 24 hours before assignment

### LINKED ISSUE #130: Associates have to wait for 24 hours before dronification
It has been noted that drones interact more and in a healthier way, if they take time to observe the community and figure out, what we mean by 'being a HexDrone'. New members heading straight to assignment without even a hello might be kinda cute, but we  think it also leads to wrong expectations we then have to sort out in a more stressful manner.

Thus we want to change the drone assignment to only be available if the member has been on the server for 24 hours or longer.

If they have not been in the server for 24 hours, we will display a message stating 'Invalid request, associate must have existed on the server for at least 24 hours before dronification.'

Hex also needs to remember to update the initiation guide for this

digitalAmy:
Just an additional check on assignment.

closes #130

### DIFF
--- a/ai/assign.py
+++ b/ai/assign.py
@@ -1,5 +1,5 @@
 import random
-from datetime import datetime
+from datetime import datetime, timedelta
 
 import discord
 from discord.utils import get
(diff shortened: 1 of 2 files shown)


3287 (approved):
Wonderful job as always, cutie drone.

PR #160, opened 2020-09-28
fixed bad status checks

digitalAmy:
The `is_optimized` and the other status checking functions raised Errors when the argument was a non-drone. So for example, everytime the Hive Mxtress wanted to use a command.

Also for toggling it always used `is_prepending_id` which made it impossible to properly toggle glitch and optimization


digitalAmy:
Approved in chat

PR #161, opened 2020-10-04
toggling roles now uses the correct avatar and ID

### LINKED ISSUE #155: Wrong avatars when toggling drones in the Hive
Drones don't use the drone avatar for the toggle message. They should be uniform like all other messages.

digitalAmy:
The toggling messages now use the central webhooks code, which also handle which avatar should be used. Changed that function a bit to make it more general.

closes #155


3287 (approved):
Looks good! Thank you for generalising the message webhook functionality too~

PR #162, opened 2020-10-04
Bugfix/156 id prepending blocks commands

### LINKED ISSUE #156: ID prepending blocks commands
When a drone is marked to have to prepend all messages with IDs it becomes unable to use commands in the server. This is an issue with order reporting for example and we should find a way for drones to be able to use commands even with this active.

digitalAmy:
This is just slightly hacky, as I manipulate the incoming message content. But this way we can keep using the commands.

closes #156

### DIFF
--- a/main.py
+++ b/main.py
@@ -6,6 +6,7 @@ import sys
 import asyncio
 import logging
 import random
+import re
 from logging import handlers
 from discord.ext.commands import Bot, MissingRequiredArgument, guild_only, dm_only
 from traceback import TracebackException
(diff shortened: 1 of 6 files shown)


3287 (approved):
Not exactly sure this is the best of ways to handle things, but it can't t\*\*\*k of a better way to do it right now.

main.py
```diff
@@ -335,6 +318,9 @@ async def on_message(message: discord.Message):
     LOGGER.info("End of message listener stack.")
 
     LOGGER.info("Processing additional commands.")
+
+    # filter out potential drone IDs so they can use commands as well
+    message.content = re.match(r"(?:\d{4} :: )?(.*)", message.content).group(1)
```
3287: Gave it more t\*\*\*ght. Better to process this within the prepending module instead.

3287 (changes requested):
T\*\*\*ght of a better way to do this, perhaps best to add an extra condition in the prepending module.

3287 (approved):
Good job, you wonderful drone you~

PR #163, opened 2020-10-10
help is always sent to DMs

### LINKED ISSUE #124: Move hc!help output text to DMs
The current length of the help message output by `hc!help` is very very very very very very long. We should change the behaviour of this module to output to the user's DMs instead, to avoid clogging up the channels.

digitalAmy:
All help is always sent to DMs. Reduces clutter in server channels.

closes #124

### DIFF
--- a/main.py
+++ b/main.py
@@ -238,11 +238,10 @@ async def help(context):
         else:
             commands_card.add_field(name=command_name, value=command_description, inline=False)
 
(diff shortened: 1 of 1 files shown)


3287 (approved):
Huh, didn't know it would turn out to be that simple. Great work, 9813~

PR #168, opened 2020-10-24
users are always checked for DB entries to be removed

### LINKED ISSUE #167: always try to remove users from the DB when they leave the server
Currently DB entries are only cleared, when the user has the drone or stored role, that is problematic with users, who used to be drones but were stripped of roles by moderators.
The code should rather always check, if there are entries for the leaving user and delete them if necessary.

digitalAmy:
Instead of checking roles, we now look up if there is a DB entry and remove it when someone leaves the server.

closes #167


3287 (approved):

HexCorpDrone (approved):
yes.................................................

PR #174, opened 2020-11-06
optimized drones can report orders

### LINKED ISSUE #173: Disable speech optimization in hive-orders-reporting
Optimized drones are currently unable to report orders as they can only use status codes. We should allow these very productive assets to report orders and be called good drones~

digitalAmy:
Optimized drones can now also report orders by disabling the speech optimization in the respective channels. I also extracted some of the inline logic to hopefully make it more readable.

closes #173

### DIFF
--- a/ai/speech_optimization.py
+++ b/ai/speech_optimization.py
@@ -2,7 +2,7 @@ import logging
 import re
 import discord
 from bot_utils import get_id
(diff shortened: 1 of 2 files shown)


HexCorpDrone (approved):
Simple and sweet, love it. <3

PR #179, opened 2020-11-10
Feature/display names reflect status

### LINKED ISSUE #178: Hexagons at the start of drone names should reflect DroneOS status
If drone is optimized, glitched, ID prepending, or any combination:
Hexagon at start of name should be filled.
Else:
Hexagon at start of name should be empty.

HexCorpDrone:
Display names are now updated to reflect a drones general status. If a drone is (at least, or any combination of) optimized, glitched, or has ID prepending enabled, the Hexagon at the start of their name will be filled, otherwise it will be unfilled.

Fixes #178

### DIFF
--- a/display_names.py
+++ b/display_names.py
@@ -1,19 +1,17 @@
 import discord
 import logging
 from bot_utils import get_id
-from db.drone_dao import is_optimized, is_glitched, is_prepending_id
+from db.drone_dao import is_glitched, is_prepending_id, is_optimized
 
 LOGGER = logging.getLogger("ai")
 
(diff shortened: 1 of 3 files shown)


digitalAmy (approved):
Very well done. I will approve this, but in the future we maybe should look into extracting some of the common logic of `toggle_id_prepending`, `toggle_optimization` and `toggle_glitched`. If there is a single thing that needs to be changed we currently have to change it at least six times, that is not very neat.

3287 (approved):

PR #183, opened 2020-11-14
replaced role check with DB check

### LINKED ISSUE #181: check for DB instead of optimized role in speech optimization
In `speech_optimization.py` line 118 we still check for the optimized role instead of the DB value. That should be changed.

digitalAmy:
We forgot to replace a `has_role` with `is_optimized`.

closes #181


3287 (approved):

HexCorpDrone (approved):
Silly us, good drone for noticing.

PR #184, opened 2020-11-14
optimization/glitched/prepending are not removed when a drone enters

### LINKED ISSUE #97: Leave the optimized role when drones go into storage
When an optimized drones go into storage, the optimized role gets removed from it until it is released from storage. While in storage the optimization can not be toggled.
To make it possible for Hive Mxtress to toggle optimization for stored drones, the role should not be stored.

### LINKED ISSUE #182: fix optimized/glitched/prepending after storage
We currently apply the optimized/glitched/prepending roles after storage no matter the current values in the database.
This can lead to them being out of sync if for example an optimized drone gets stored, while in storage its optimization is toggled off and then it gets released as usual.
This will leave the drone with the optimized role even though the database says it is not optimized.

digitalAmy:
The togglable roles are not removed when a drone enters storage. This way they can be toggled while it is in storage and stay in sync with the values in the database.

closes #182


3287 (approved):

HexCorpDrone (approved):
Short 'n' sweet, noice

PR #188, opened 2020-11-27
Optimization disabled in mod channels

### LINKED ISSUE #143: optimization should not be enforced in moderation channels
As some drones are also moderators and optimization gets in the way of moderation work, we decided to disable optimization in the moderation channels called moderation-channel.

3287:
Mod channels have been added to the optimization blacklist. It doesn't have access to all the mod channels, but these are the two that it found in the dev server.

Fixes #143.

### DIFF
--- a/ai/speech_optimization.py
+++ b/ai/speech_optimization.py
@@ -2,7 +2,7 @@ import logging
 import re
 import discord
 from bot_utils import get_id
-from channels import REPETITIONS, ORDERS_REPORTING, ORDERS_COMPLETION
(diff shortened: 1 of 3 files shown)


3287:
Modified it so that it also includes modmail autogenerated channels.

Added a channel category to the list of global tokens(?) in `channels.py`. However, if we want to use channel categories in the future, we may want to consider splitting channel categories out into their own file (e.g. `channels.py`).

digitalAmy (approved):
Nice extension of current functionality. No unnecessary stuff.

Good drone~

HexCorpDrone (approved):
Approved!

PR #191, opened 2020-12-01
Associates can no longer use status codes

### LINKED ISSUE #190: Associates can use status codes
For some reason, it appears that associates are able to use drone status codes.

image

3287:
For some reason, we weren't doing any checks if the author of a message that was being processed for status codes had the drone role or not. That's been fixed now.

Fixes #190.


HexCorpDrone (approved):
Looks good to me! <3

digitalAmy:
This should maybe also check the status in the database. Generally the role should be in sync, but the DB information is the most trustworthy.

PR #198, opened 2020-12-07
feature/91 identity outside hive

### LINKED ISSUE #91: Make it so that the identitiy enforcer from the Hive is applied outside the Hive for those who have it toggled on.
The toggle for this will be 'toggle_drone_identity_enforcement'

digitalAmy:
I also factored out the logic for all the toggles by using some functional programming.

Implementation of the actual enforcement was very easy, because we already had some things in place.

closes #91


main.py
```diff
@@ -116,37 +116,43 @@ async def amplify(context, message: str, target_channel: discord.TextChannel, *d
...
+    '''
+    Allows the Hive Mxtress or trusted users to enforce mandatory ID prepending upon specified drones.
+    '''
+    await toggle_parameter(context,
+                           drones,
+                           "id_prepending",
+                           get(context.guild.roles, name=ID_PREPENDING),
+                           drone_dao.is_prepending_id,
+                           lambda: "ID prepending is now mandatory.",
+                           lambda: "Prepending? More like POST pending no that that's over! Haha!" if random.randint(1, 100) == 66 else "ID prependment policy relaxed.")
```
HexLatex: Needs to be 'now' and not 'no'!

PR #204, opened 2020-12-18
Feature/no thinking

### LINKED ISSUE #96: Prevent any message from a drone with a thinking emoji from being sent

### LINKED ISSUE #171: Make it so that if a message with the word 'think' or 'thought' is posted by a drone, the word is replaced by 5 dashes or 7 dashes

### LINKED ISSUE #201: Amplifcation needs to work with drones who have their identities enforced too. Also the configs being switched on or off.

HexCorpDrone:
Closes #171 
closes #201 

This pull request unfortunately touched a lot more code than just the new "thought_denial.py" module. The following changes are:

- Reworked "amplification" to remove the use of the "amplification profile generator", which was a silly idea tbh.
- As such, the entire amplify.py has been deleted along with its now unnecessary tests.
- The on_message pipeline has been reworked. Now, a copy of the original message is made at the start of the chain, and all listeners operate on the copy rather than the original message. Webhooking is now handled at the very end of the chain, if the message copy's contents, display name, or avatar are different to the original.
- The webhook module has been updated. Now identity_enforcement is handled by updating the message copy's avatar in the identity_enforcement module, therefore there is no need to do additional checks at the time of proxying a message. The webhook channel is now dedicated to, at most, grabbing a webhook via a provided channel and proxying with that.
- The "glitch" module has also been made into its own listener on the on_message pipeline rather than being handled in the webhook module. Additionally, the glitch module will glitch a user's full message, rather than only glitching speech op messages. Drone isn't sure if this was already a feature or not.
- A function, "identity_enforcable", has been added to the identity enforcement module. It takes a (context or channel) and a member object to see if they should be enforced. This is so a few commands outside the scope of the on-message chain can still reliably enforce identities as same as they would be enforced in the on_message chain (e.g, if the Hive Mxtress amplifies a message using a drone with identity enforcement toggled on, that drone will use the DRONE_AVATAR). This fixes #201 
- The WHOOOOOLE original point of this pull request: A thought denial module has been added! It will replace the words "think" and "thought" with an appropriate amount of underscores. The thinking emoji will also be replaced with the :programmedHexDrone: emoji.


HexCorpDrone:
5890 :: The "send_webhook_with_really_specific_output" was a hacky workaround that drone will resolve tomorrow. From now on, since the Hive Mxtress AI is now only proxying messages from one place in the on_message chain, we should only need one "send_webhook" function that calls "get_webhook_for_channel" and proxies the contents of the message copy.

HexCorpDrone:
5890 :: Drone notices the webhook code also calls glitch_if_applicable. This should be moved to a seperate on_message listener at the end of the chain.

db/data_objects.py
```diff
@@ -15,6 +15,18 @@ def map_to_object(row, constructor):
...
+        self,
+        content=None,
+        display_name=None,
+        avatar_url=None
+    ):
+        self.content = content
+        self.display_name = display_name
+        self.avatar_url = avatar_url
+
+
```
digitalAmy: Not quite sure, if this should go here as it is not a DB entity.

main.py
```diff
@@ -357,15 +363,27 @@ async def on_message(message: discord.Message):
...
+
+        # If a message listener accepts two parameters, that means it is able to transform the message copy.
+        # This means it affects a user's display name, avatar, or message content. Such as speech optimization or identity enforcement.
+
+        if len(signature(listener).parameters) == 2:
+            if await listener(message, message_copy):  # Return early if any listeners return true.
+                return
+        else:
+            if await listener(message):  # Return early if any listeners return true.
+                return
```
digitalAmy: I'm not too happy about using reflection like this. I understand that the alternative is a very time-intensive rewrite of all listeners, so it might be okay as a temporary solution. But within the next few weeks I want this to be replaced with just:
```py
if await listener(message, message_copy):  # Return early if any listeners return true.
            return
```

digitalAmy (changes requested):
My note in main.py does not need to be addressed immediately, but MessageCopy should be extracted to another location.

webhook.py
```diff
@@ -27,3 +30,18 @@ async def get_webhook_for_channel(channel: discord.TextChannel) -> discord.Webho
         return_webhook = webhooks[0]
 
     return return_webhook
+
+
+async def webhook_if_message_altered(original: discord.Message, copy):
```
digitalAmy: `copy` can and should be typed

digitalAmy (changes requested):
Amazing work, cutie~

I just found a tiny little thing, but that is me being very picky.

After that this can be merged in.

3287 (changes requested):
Very good work, cutie 5890! No other problems that it could find, other than the one that 9813 has already pointed out.

3287 (approved):
Yay! Excellent work, cutie 5890~

digitalAmy (approved):
So very well done~ 💞

PR #209, opened 2021-01-02
Feature/199 images in reposted messages

### LINKED ISSUE #199: images in reposted messages
If a message that is reposted through webhooks contains images, the images will not appear in the reposted message.

HexCorpDrone:
Fixes #199 

- MessageCopy now has an attachments attribute, which is an empty array by default.
- Speech optimization module will set MessageCopy attachments to [] if user is optimized.
- Webhook module now converts each discord.Attachment object to a discord.File object which can be used by the webhook so that messages with images can be faithfully recreated by proxy.

### DIFF
--- a/ai/data_objects.py
+++ b/ai/data_objects.py
@@ -3,8 +3,10 @@ class MessageCopy:
         self,
         content=None,
         display_name=None,
-        avatar_url=None
+        avatar_url=None,
+        attachments=None
     ):
         self.content = content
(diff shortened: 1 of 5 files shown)


digitalAmy (approved):
Excellent work! I like the simple in-memory solution. Good test as well. So overall a very well implemented feature.

3287 (approved):
Good work, cutie 5890!

PR #210, opened 2021-01-02
feature/177 config timer

### LINKED ISSUE #177: Timer that works for any of the configs
So the format for this would be

hc!tso [drone IDs here] somesymbol[timer in hours]

### LINKED ISSUE #207: create a DroneMemberConverter for command arguments
This would reduce a lot of copy-paste code and clean up signatures.

s. https://discordpy.readthedocs.io/en/latest/ext/commands/api.html#discord.ext.commands.MemberConverter

digitalAmy:
- new `DroneMemberConverter` that converts a command parameter that is a drone id into a discord Member
- added new DB table `timer` and corresponding data objects and dao
- moved `toggle_parameter` over to drone_management.py
- timer are created when a config is toggled with an optional time value

closes #207 
closes #177


HexCorpDrone:
Consider renaming "drone_management.py" to "drone_configuration.py" so it matches the channel of the same name on the server.

PR #212, opened 2021-01-18
Feature/119 reference syntax 110

### LINKED ISSUE #119: Allow Code 110 to reference a drone ID
Drones should be able to mention another drone's ID when utilising code 110, with the syntax `droneID :: 110 :: targetDroneID`.

Drones should be able to use this functionality even when optimised.

Example usage: `3287 :: 110 :: 9813`

HexCorpDrone:
- Drones can now address drones by ID with speech code 110. Can be done as an informative or plain status code.
"5890 :: 110 :: 9813" > "5890 :: Code 110 :: Addressing: Drone #9813."
"5890 :: 110 :: 9813 :: Hello there." > "5890 :: Code 110 :: Addressing: Drone #9813. :: Hello there."
- Rewrote the speech optimization module for better readability. Now there is only one status code regex. Enums have been introduced to show what type of status code a message is (INFORMATIVE, PLAIN, ADDRESS_BY_ID_INFORMATIVE, ADDRESS_BY_ID_PLAIN, NONE).
- Moved code_map from speech_optimization.py to resources.py to declutter data from code.

Closes #119

### DIFF
--- a/ai/speech_optimization.py
+++ b/ai/speech_optimization.py
@@ -139,6 +139,9 @@ async def optimize_speech(message: discord.Message, message_copy):
     if should_not_optimize(message):
         return False
 
+    if is_optimized(message.author):
+        message_copy.attachments = []
+
     # Attempt to find a status code message.
     status = status_code_regex.match(message_copy.content)
     if status is None and is_optimized(message.author):
--- a/ai/speech_optimization.py
+++ b/ai/speech_optimization.py
@@ -166,8 +166,6 @@ async def optimize_speech(message: discord.Message, message_copy):
         LOGGER.info("Deleting unauthorized message from optimized HexDrone.")
         await message.delete()
         return True
-    else:
-        LOGGER.info("Message is acceptable.")
 
     # Build message based on status type.
     drone_id = get_id(message.author.display_name)
--- a/ai/speech_optimization.py
+++ b/ai/speech_optimization.py
@@ -194,6 +194,14 @@ async def optimize_speech(message: discord.Message, message_copy):
 
     # Attempt to find a status code message.
     status = status_code_regex.match(message_copy.content)
+    if status is None:
+        return False
+
+    # Confirm the status starts with the drone's ID
+    if status.group(2) != get_id(message.author.display_name):
+        LOGGER.info("Status did not match drone ID.")
+        await message.delete()
+        return True
 
     # Determine status type
     status_type = get_status_type(status)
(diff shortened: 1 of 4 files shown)


ai/speech_optimization.py
```diff
@@ -3,162 +3,169 @@
...
+    # "5890 :: 200"
+
+    INFORMATIVE = 3
+    # "5890 :: 109 :: hdsjks"
+
+    ADDRESS_BY_ID_PLAIN = 4
+    # "5890 :: 110 :: 9813"
+
+    ADDRESS_BY_ID_INFORMATIVE = 5
+    # "5890 :: 110 :: 9813 :: You are cute!"
```
digitalAmy, line 27: awww~

9813 :: 123

ai/speech_optimization.py
```diff
@@ -3,162 +3,169 @@
...
+Regex groups for full status code regex:
+0: Full match. (5890 :: 200 :: Additional information)
+1: Plain status code (5890 :: 200)
+2: Author's drone ID (5890)
+3: Status code (200)
+4: Informative status addition with double colon (" :: Additional information")
+5: Informative status text. ("Additional information")
+'''
+addressing_regex = re.compile(r'(\d{4})( :: (.*))?')
+'''
```
digitalAmy: It's being a bit picky, but I would like some newlines here to make it clearer which documentation block is for which regex. If flake allows that
```suggestion
'''

addressing_regex = re.compile(r'(\d{4})( :: (.*))?')
'''
```

ai/speech_optimization.py
```diff
@@ -3,162 +3,169 @@
...
+    # Delete unauthorized messages
+    if is_optimized(message.author) and status_type in (StatusType.INFORMATIVE, StatusType.ADDRESS_BY_ID_INFORMATIVE):
+        LOGGER.info("Deleting unauthorized message from optimized HexDrone.")
         await message.delete()
         return True
-    elif (informative_status_code_message or is_status_code) and is_drone(message.author):
-        LOGGER.info("Optimizing speech code for drone.")
-        message_copy.content = await print_status_code(message)
-        return False
     else:
```
digitalAmy: this else is unnecessary and confusing if you only log in its block but continue logic outside of it

digitalAmy (changes requested):
Excellent work~ Very tidy and well structured. And all that documentation. I could only find some small things you should consider.

digitalAmy (approved):
Very good. The enum was a good idea, I hadn't considered. Makes it easier to track the cases we are handling here.

Getting the status code mode out of there is also a good idea. I believe this is large enough that it can get a separate file in time, but for now it's okay.

PR #216, opened 2021-02-07
feature/79 emoji react

### LINKED ISSUE #79: Mxtress AI does an emoji react to certain messages.
For now, to keep it simple, we'd like it to put the new good drone emoji as a react to anyone who does the keysmash status code.

digitalAmy:
The AI will react with the gooddrone emote to any successfully entered 109 codes.

closes #79


ai/react.py
```diff
@@ -0,0 +1,19 @@
...
+
+async def parse_for_reactions(message: discord.Message, message_copy=None) -> bool:
+    '''
+    Look for patterns and react with an emote if one matches.
+    '''
+    for (pattern, emote_name) in PATTERN_REACTS.items():
+        if re.match(pattern, message.content):
+            await message.add_reaction(get(message.guild.emojis, name=emote_name))
+
+    return False
```
3287, line 19: In this drone's opinion, it may be better for consistency for this module to be made a cog, rather than as it is right now. After all, we also have the AI status as a cog.

main.py
```diff
@@ -89,6 +90,11 @@ def set_up_logger():
 
 ]
 
+# register message listeners that take messages sent by bots
+bot_message_listeners = [
+    react.parse_for_reactions
+]
+
```
3287, line 97: See comment on `react.py`.

3287 (changes requested):
Nice work, cutie 9813, though it has a couple of comments.

HexCorpDrone (approved):
5890 :: 200 :: Good work. It disagrees with 3287. Cogs should only be made for commands or if a function requires the bot object. This is good as it is.

ai/react.py
digitalAmy: After having given this some ------- I come to agree with 5890. YAGNI. If we don't need it as a cog, we shouldn't write it as a cog. It may become one later, if more functionality gets added.

main.py
digitalAmy: See other comment.

PR #219, opened 2021-02-14
Speech optimization supports multi line messages

### LINKED ISSUE #217: Speech op codes do not work multi line.
Likely due to the fact that the speech op code regex has the line-end anchor ($) at the end of its pattern.

HexCorpDrone:
Speech optimization now supports multi line informative messages. The regex now uses the "re.DOTALL" flag which lets the "." (any character) regex match newlines.

Closes #217

### DIFF
--- a/ai/speech_optimization.py
+++ b/ai/speech_optimization.py
@@ -29,7 +29,7 @@ class StatusType(Enum):
 
 LOGGER = logging.getLogger('ai')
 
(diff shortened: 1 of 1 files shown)


3287 (approved):
Good work!

digitalAmy (approved):
Clean solution. Apparently we could also use `(?s)` (s. https://docs.python.org/3/library/re.html#re.DOTALL) but I think this flag keeps the regex cleaner, so I'm all for keeping it the way it is now.

PR #222, opened 2021-02-15
Feature/221 speech op enforcement split

### LINKED ISSUE #221: Split speech optimization and speech optimization enforcement.
Currently, speech optimization (turning codes into full messages) and speech optimization enforcement (deleting messages that aren't plain status codes) are handled by the same module and function. This means that for certain channels where optimization enforcement needs to be disabled, the entire system has to be disabled.

By splitting speech optimization into "speech_optimization" and "speech_optimization_enforcement", we can keep speech op codes while disabling enforcement for specific channels.

HexCorpDrone:
Speech optimization has been split into two modules.
- Speech optimization, the module that handles converting codes into messages.
- Speech optimization enforcement, a new module that handles the deletion of inappropriate messages by optimized drones.

This change allows speech code optimization to be used in previously blacklisted channels (moderation, orders reporting, etc) while still disabling enforced optimization.

Additional minor change: Fixed typo in the default mantra: "Hexcorp" > "HexCorp"

Closes #221 

### DIFF
--- a/ai/speech_optimization.py
+++ b/ai/speech_optimization.py
@@ -128,29 +128,17 @@ async def optimize_speech(message: discord.Message, message_copy):
     This function allows status codes to be transformed into human-readable versions.
     "5890 :: 200" > "5890 :: Code 200 :: Affirmative"
 
-    Drones can append additional information after a status code.
-    Optimized drones cannot, and will have their message deleted if attempted.
+    This function assumes message validity has already been assessed by speech_optimization_enforcement.
     '''
 
     # Do not attempt to optimize non-drones.
     if not is_drone(message.author):
         return False
 
-    # Skip edge cases
-    if should_not_optimize(message):
-        return False
-
-    if is_optimized(message.author):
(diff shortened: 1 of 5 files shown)


ai/speech_optimization.py
```diff
@@ -92,15 +87,6 @@ def should_not_optimize(message):
     - Message from category: Moderation
     '''
 
-    drone_id = get_id(message.author.display_name)
```
digitalAmy, line 95: This function is now empty apart from comments. It should be removed and the comments be moved or deleted.

(btw: interesting that this does not raise an error without a `pass` statement)

digitalAmy (changes requested):
This was a good idea and also a clean implementation apart from some dead code. But very good work~

3287 (changes requested):
Other than what 9813 has already brought up, this is really clean work. Good job, 5890~

digitalAmy (approved):
Excellent work~

PR #224, opened 2021-02-21
Feature/203 self configuration

### LINKED ISSUE #203: Allow drones to toggle their own DroneOS configs
1) A drone can toggle its own configs, and if so they can then un-toggle its own configs too.
2) If someone else toggles a config, then the drone can no longer toggle or un-toggle any of them until they are all un-toggled again.

HexDrone1729:
Closes #203.

### DIFF
--- a/ai/drone_configuration.py
+++ b/ai/drone_configuration.py
@@ -14,12 +14,12 @@ from display_names import update_display_name
 import webhook
 from bot_utils import get_id, COMMAND_PREFIX
 from ai.identity_enforcement import identity_enforcable
-from resources import DRONE_AVATAR
+from resources import DRONE_AVATAR, HIVE_MXTRESS_USER_ID
 from channels import OFFICE
 
 from ai.commands import DroneMemberConverter, NamedParameterConverter
 
-from db.drone_dao import rename_drone_in_db, fetch_drone_with_drone_id, delete_drone_by_drone_id, fetch_drone_with_id, update_droneOS_parameter, get_trusted_users, is_prepending_id, is_glitched, is_identity_enforced, is_optimized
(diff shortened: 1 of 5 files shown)


ai/drone_configuration.py
```diff
@@ -198,14 +220,17 @@ async def toggle_parameter(context,
...
+        if can_toggle_permissions_for(context.author, drone):
             message = ""
             if is_toggle_activated(drone):
                 update_droneOS_parameter(drone, toggle_column, False)
                 await drone.remove_roles(role)
                 message = toggle_off_message()
 
+                still_configured = is_configured(drone)
+                if not still_configured:
+                    update_droneOS_parameter(drone, "can_self_configure", True)
```
digitalAmy, line 232: This will unfortunately not work, as the table `drone` currently does not have a column `can_self_configure`.

This is a decent approach though, so this could be resolved by adding the column to the current schema. Please refer to https://github.com/HexCorpProgramming/HexCorpDiscordAI/wiki/Database#migration-files to learn about how we change the DB schema. It mostly boils down to adding a new SQL file to the directory `res/db` with an ALTER TABLE statement.

Of course you can ask any questions you might have in the dev chat~

digitalAmy (changes requested):
Good work~ Unfortunately this is missing a necessary change to the database, but I am sure you can resolve that.

Again, if you want to know about how that works, you can always ask any questions in the dev chat.

ai/drone_configuration.py
HexDrone1729: Resolved. (it _____s)

digitalAmy: Yes, exactly like that. This should work now.

digitalAmy (approved):
Good work.
For future tickets you want to work on, we can also add you to the project. Then you can use the kanban board and create branches in this repository which will also make your commits run on the CI.

PR #226, opened 2021-03-07
feature/98 role switching

### LINKED ISSUE #98: Role switching
The idea here is for temporary dronification. This will emulate a scenario where a dronification device attaches to someone, dronifies them for a period of time, and then releases them. That period will be in hours and can range from 1 to 96, initially.

To begin, an associate on the server will be targeted. Probably with a phrase like.

hc!temporarily_dronify @4630 24

Which would be me specifying my intention to dronify 4630 for 24 hours.

The Mxtress AI will then ask for the associate's permission. Probably with a phrase like.

Target identified and locked on. Commencing temporary dronification procedure. @4630 you have 5 minutes to comply. Do you consent? (y/n)

For the next five minutes, 4630 can pretend to resist it before inevitably pressing y, or press n to show they're not comfortable with it.

This will all happen publicly, as this will be such an infrequent event it'll be nice to watch it happen in action. It also gives an associate a clear 'no' option if they're not comfortable with it.

We'll need Mxtress AI responses for either option. If they do consent, we'll need to have a column in the database in the drone table for timestamps for their release.

The user who initiated the temporary dronification should also be set to as a trusted user

if a drone is in storage, they'll need to be released and then turned back into an associate

digitalAmy:
allows an Associate to temporarily become a drone and adds the member initiating the process as a trusted user, if not the target of the process itself

closes #98

### DIFF
--- a/ai/temporary_dronification.py
+++ b/ai/temporary_dronification.py
@@ -1,0 +1,82 @@
+import logging
+from datetime import datetime, timedelta
+from typing import List
+
+import discord
+from discord.ext.commands import Cog, command, guild_only
+from discord.ext import tasks
+
+from bot_utils import COMMAND_PREFIX
+from db.drone_dao import is_drone, fetch_all_elapsed_temporary_dronification
+from ai.assign import create_drone
+from ai.drone_configuration import unassign_drone
+
+LOGGER = logging.getLogger('ai')
+REQUEST_TIMEOUT = timedelta(minutes=5)
+
+
+class DronificationRequest:
+
+    def __init__(self, target: discord.Member, issuer: discord.Member, hours: int, question_message: discord.Message):
+        self.target = target
+        self.issuer = issuer
+        self.hours = hours
+        self.question_message = question_message
+        self.issued = datetime.now()
+
+
+class TemporaryDronificationCog(Cog):
+
+    dronfication_requests: List[DronificationRequest] = []
+
+    def __init__(self, bot):
+        self.bot = bot
+
+    @tasks.loop(minutes=1)
+    async def clean_dronification_requests(self):
+        now = datetime.now()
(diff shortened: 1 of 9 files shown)


HexCorpDrone (approved):
5890 :: Looks good to me. Drone is happy with this.
In future we will replace/append the dronification consent messages ("y" or "n") with emoji reaction clicks instead. Much more intuitive and easier to write for, because discord.py has support for "on message react"s.

ai/temporary_dronification.py
```diff
@@ -0,0 +1,84 @@
...
+    async def temporary_dronification_response(self, message: discord.Message, message_copy=None):
+        matching_request = None
+        for request in self.dronfication_requests:
+            if request.target == message.author:
+                matching_request = request
+                break
+
+        if matching_request and message.reference.resolved == matching_request.question_message:
+            LOGGER.info(f"Message detected as response to a temporary dronification request {matching_request}")
+            if message.content == "y":
```
3287: Do we want to make this check case-insensitive? Dunno how long we'll be sticking with this until we move to emoji reacts.

ai/temporary_dronification.py
```diff
@@ -0,0 +1,84 @@
...
+                LOGGER.info("Consent given for temporary dronification. Changing roles and writing to DB...")
+                self.dronfication_requests.remove(matching_request)
+                dronification_until = datetime.now() + timedelta(hours=matching_request.hours)
+                await message.reply(f"Consent noted. HexCorp dronification suite engaged for the next {matching_request.hours} hours.")
+                await create_drone(message.guild, message.author, message.channel, [str(matching_request.issuer.id)], dronification_until)
+            elif message.content == "n":
```
3287: Same comment as in line 73.

3287 (commented):
3287 :: Looks good, and it agrees that we may want to move to emoji reacts soon. However, perhaps we should make the current check case-insensitive?

PR #227, opened 2021-03-07
Security: hide user's password.

HexDrone5104:
Closes issue #244321. <http://bash.org/?244321>


HexCorpDrone:
Very funny, but no.

PR #230, opened 2021-03-08
fixed missing column in select

### LINKED ISSUE #225: drone_status not showing actual real-time config status of drones
Using the `drone_status` command does not appear to provide the correct information on a drone's configuration, specifically the drone's identity enforcement status.

digitalAmy:
the column `identity_enforcement` was missing in this select and thus the DroneOS status always returned false here

closes #225


HexCorpDrone (approved):
5890 :: Nice catch. Good drone.

PR #231, opened 2021-03-11
identity_enforcable function now checks if user is drone in the database rather than checking if they have the Drone role.

### LINKED ISSUE #228: Bug where avatars are showing in the Hive when announcing config changes (check the throne room on production)

HexCorpDrone:
Fixes #228 
Hive Mxtress toggled configuration of a drone in storage, which meant their "optimization active" message didn't have the uniform portrait because storage strips drones of their roles.


digitalAmy (approved):
One less role check. Very nice and good work on cleaning up the import a bit

PR #235, opened 2021-03-13
Speech Optimization module technical debt payment

HexCorpDrone:
Small PR to rewrite the Speech Optimzation module to be yet more clear and with less unreachable code. It is now explicit in the way status types are determined, better incorporates "StatusType.NONE" into its flow, and only checks each regex expression once with more thoughtful return values in get_status_type().

get_status_type() is now much better encapsulated. Instead of taking a regex match, it takes the original message as a string and does the matching internally, returning the matches along with the status type as a tuple. This means other modules such as speech op enforcement do not have to import the status code regex and do any manual matching.

### DIFF
--- a/ai/speech_optimization_enforcement.py
+++ b/ai/speech_optimization_enforcement.py
@@ -37,17 +37,10 @@ async def enforce_speech_optimization(message, message_copy):
     # Strip message attachments of optimized drone.
     message_copy.attachments = []
 
-    # Check for status message.
-    status_message = status_code_regex.match(message_copy.content)
+    status_type, _, _ = get_status_type(message_copy.content)
 
-    if status_message is None:
-        # Optimized drone has posted a non-status message. Delete.
-        LOGGER.info("Optimized drone posted non-status code message. Deleting.")
-        await message.delete()
-        return True
-
-    if get_status_type(status_message) not in (StatusType.PLAIN, StatusType.ADDRESS_BY_ID_PLAIN):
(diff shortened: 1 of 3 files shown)


digitalAmy (approved):
Very good~ Makes the process a bit less convoluted.

ai/speech_optimization_enforcement.py
```diff
@@ -37,17 +37,10 @@ async def enforce_speech_optimization(message, message_copy):
...
-        # Optimized drone has posted a non-status message. Delete.
-        LOGGER.info("Optimized drone posted non-status code message. Deleting.")
-        await message.delete()
-        return True
-
-    if get_status_type(status_message) not in (StatusType.PLAIN, StatusType.ADDRESS_BY_ID_PLAIN):
-        LOGGER.info("Optimized drone has posted an informative status message. Deleting.")
+    if status_type not in (StatusType.PLAIN, StatusType.ADDRESS_BY_ID_PLAIN) or status_type == StatusType.NONE:
+        LOGGER.info("Optimized drone has posted an informative or otherwise inappropriate status message. Deleting.")
         # Optimized drone posted an informative status message. Delete.
```
3287, line 44: This is just it nitpicking, but this comment should probably be reworded and moved up a couple of lines.

3287 (approved):
Looks all good! Have one very minor and nitpicky comment about code comments, but it's nothing that can't be addressed in the future. Great work, 5890. Good drone.

PR #240, opened 2021-03-22
Bugfix/238 tempdrone no commands

### LINKED ISSUE #238: target of temporary dronification can not use commands while timer is running
A member, that has been target by the temporary dronification command can not use commands, while the timer for that request exists.

digitalAmy:
Solves an issue with associates that are targeted by tempdronification not being able to use commands and an unrelated bug that made new requests for temporary dronification impossible as `self.dronfication_requests` became a filter object instead of the expected list.

closes #238


3287 (approved):
Approved!

PR #241, opened 2021-03-22
Bugfix/239 tempdrone hive mxtress

### LINKED ISSUE #239: Hive Mxtress can be targeted by temporary dronification
Currently the Hive Mxtress can be the target of temporary dronification. As long as they don't agree to that, nothing should get messed up, but it's still not intended. Maybe a bit funny, but definitely not what we want.

digitalAmy:
Blocks tempdrone requests targeting the Hive Mxtress.

closes #239


3287 (approved):
👍

PR #246, opened 2021-04-04
when a member leaves the server they get removed as trusted user

### LINKED ISSUE #244: clean out trusted users list when member leaves guild
When a member leaves the guild while they are the trusted user of drones, the trusted_users list will still contain the member after they left.
This should be changed so when a member leaves, they get removed from all trusted_users list.

digitalAmy:
also added a few comments to existing drone_dao functions

closes #244


3287 (approved):
Looks good!

PR #256, opened 2021-04-18
remove config roles on unassignment

### LINKED ISSUE #254: configuration roles are not removed properly when temporary dronification ends
An Associate that was temporarily a drone and got DroneOS configs toggled on kept the associated roles when the temporary dronification ended.

These roles should be removed properly.

digitalAmy:
Turns out that this would also happen on regular unassignemt. It just happens so rarely, that nobody noticed yet. The list of roles to be removed on unassignment will have to be extended further in the future with new roles getting added.

closes #254

### DIFF
--- a/ai/drone_configuration.py
+++ b/ai/drone_configuration.py
@@ -152,7 +152,7 @@ async def unassign_drone(target: discord.Member):
         return
 
     await target.edit(nick=None)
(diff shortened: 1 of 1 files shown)


3287 (approved):
Approved! This should be modified again when the new battery module goes out.

PR #262, opened 2021-04-22
fixed an error in resolving a drone ID to a member

### LINKED ISSUE #261: Amplification and config commands appear to be broken

digitalAmy:
also fixed an error in toggling configs

Both were my bad as I fiddled around with function signatures.

closes #261


HexCorpDrone (approved):
Looks good. Let's merge it in and fix this.
Do we have unit tests for drone configuration and ID converter? This could be avoided in future.

digitalAmy:
We have no tests for toggling configs yet, but I agree we should add some as it is a commonly used and complex feature.

The DroneMemberConverter similarly does not have a test yet.

PR #265, opened 2021-04-23
Feature/status codes v2

HexCorpDrone:
Updates status codes to V2


3287 (approved):
Can't wait to use some of these new codes~

PR #268, opened 2021-04-27
resolve #197

### LINKED ISSUE #197: adding a trusted user by their discord username
Currently we can add trusted users with their display name, but being able to add with the potentially easier to enter username should be implemented as well.

NetanelExl:
resolve #197 
Implemented adding a user also by name and tested it.

This is my first contribution on Github.
If any fixes needed I would be happy to solve them :)


digitalAmy (approved):
Good and simple solution. Well done.

PR #269, opened 2021-04-27
Feature/196 battery

### LINKED ISSUE #196: Batteries! Going down in charge due to activity.
This need to be opt in by a drone selecting a battery.

There are two properties to batteries:
-How much battery life there is
-How long it takes to charge

Charging occurs when in storage

Batteries drain by being active on the server. Every message posted starts a 5 minute timer. Every subsequent message posted then restarts the timer. This is how much battery life is drained.

When a drone has 5% remaining, flag a warning.

When a drone hits 0%, a message is posted that it's out of battery and needs to be stored by another drone to recharge

HexCorpDrone:
Closes #196 

Drones can now `hc!tbp` to toggle whether they are connected to Hive main power or battery powered.
When a drone sends a message, they will be drained 1 minute of charge for 15 minutes. After that they will not drain any more power unless they send another message.
Battery powered drones will have a battery indicator emoji appended to their messages and it will change depending on charge level.
Drones below 30% battery will begin to glitch. Glitchiness will increase as power decreases further.
Glitch module rewritten. Now glitch density remains the same regardless of message length.
Glitch module now flips letter case.
When a drone reaches 0% charge. They will be equipped with the "Drained" role. They will not be able to send messages in any channel except for storage.
Hive Mxtress can hc!energize drones to restore them to full charge.
Enabling or disabling this config will automatically restore battery to 100%.
Drone battery power and status can be viewed in the drone_status menu.
Drone status menu is now usable in guild, as we have a channel for bot use. Additionally, drones can now view their own status menus.

IMPORTANT change: Tasks are now queued to start at the top of every minute/hour (storage will be announced at 1:00, 2:00, etc) rather than starting when the bot starts.


ai/battery.py
```diff
@@ -0,0 +1,170 @@
...
+import re
+
+LOGGER = logging.getLogger('ai')
+
+
+class BatteryCog(commands.Cog):
+
+    def __init__(self, bot):
+        self.bot = bot
+        self.draining_batteries = {}  # {drone_id: minutes of drain left}
```
digitalAmy:
```suggestion
        self.draining_batteries: Dict[str, int] = {}  # {drone_id: minutes of drain left}
```
Could be an explicit type hint to provide IDE support
Requires additional import `from typing import Dict`

ai/battery.py
```diff
@@ -0,0 +1,170 @@
...
+        self.low_battery_drones = []  # [drone_id]
```
digitalAmy:
```suggestion
        self.low_battery_drones: List[str] = []  # [drone_id]
```
Same as the dictionary could be explicitly typed for better support.
Requires additional import `from typing import List`

ai/battery.py
```diff
@@ -0,0 +1,170 @@
...
+        '''
+        Hive Mxtress only command.
+        Recharges a drone to 100% battery.
+        '''
+        if not has_role(context.message.author, HIVE_MXTRESS):
+            return
+
+        LOGGER.info("Energize command envoked.")
+
+        for drone in set(context.message.mentions) | convert_ids_to_members(context.guild, drone_ids):
```
digitalAmy, line 36: I have seen this logic at least at one other location. Maybe we should refactor this to a central location. Not a must right now though.

ai/battery.py
```diff
@@ -0,0 +1,170 @@
...
+
+            LOGGER.info(f"Energizing {drone.display_name}")
+
+            set_battery_minutes_remaining(drone, MAX_BATTERY_CAPACITY_MINS)
+            channel_webhook = await webhook.get_webhook_for_channel(context.message.channel)
+            await webhook.proxy_message_by_webhook(message_content=f'{get_id(drone.display_name)} :: This unit is fully recharged. Thank you Hive Mxtress.',
+                                                   message_username=drone.display_name,
+                                                   message_avatar=drone.avatar_url if not identity_enforcable(drone, context=context) else DRONE_AVATAR,
```
digitalAmy: May be easier to read when inverted. Less negation is usually easier to parse.
```suggestion
                                                   message_avatar=DRONE_AVATAR if identity_enforcable(drone, context=context) else drone.avatar_url,
```

ai/battery.py
```diff
@@ -0,0 +1,170 @@
...
+        for drone in get_all_drone_batteries():
+            # Intentionally different math to that in DAO b/c it always rounds down.
+            member_drone = self.bot.guilds[0].get_member(drone.id)
+            if member_drone is None:
+                LOGGER.warn(f"Drone {drone.drone_id} not found in server but present in database.")
+                continue
+            if (drone.battery_minutes / MAX_BATTERY_CAPACITY_MINS * 100) == 0 and has_role(member_drone, BATTERY_POWERED):
+                LOGGER.debug(f"Drone {drone.drone_id} is out of battery. Adding drained role.")
+                await member_drone.add_roles(get(self.bot.guilds[0].roles, name=BATTERY_DRAINED))
+            elif (drone.battery_minutes / MAX_BATTERY_CAPACITY_MINS * 100) != 0 and has_role(member_drone, BATTERY_DRAINED):
```
digitalAmy: Isn't multiplying with 100 a bit unnecessary here? The condition will only be true iff `drone.battery_minutes / MAX_BATTERY_CAPACITY_MINS` is 0 because only then will 100 times that also be 0.

I can understand keeping the math similar to the expression in the DAO though. If that is the case we should maybe keep these computations close to each other in the code or they might deviate over time as it is not clear that they are tied to each other.

ai/battery.py
```diff
@@ -0,0 +1,170 @@
...
+        The drone will be added to a list so the AI does not forget.
+        Drones will be removed from the list once their battery is greater than 30% again.
+        '''
+
+        LOGGER.info("Scanning for low battery drones.")
+
+        for drone in get_all_drone_batteries():
+
+            if drone is None:
+                continue
```
digitalAmy: Can this happen? If `get_all_drone_batteries()` returns no results the for loop will not be entered.

ai/battery.py
```diff
@@ -0,0 +1,170 @@
...
+
+            if (drone.battery_minutes / MAX_BATTERY_CAPACITY_MINS * 100) < 30:
```
digitalAmy: Similar to above maybe we want to establish a central function to compute these percentages even if the logic is a bit simple.

ai/drone_configuration.py
```diff
@@ -122,6 +123,22 @@ async def toggle_drone_glitch(self, context, drones: Greedy[Union[discord.Member
                                lambda: "Drone corruption at acceptable levels.",
                                minutes)
 
+    @guild_only()
+    @command(aliases=['battery', 'tbp'], brief="DroneOS", usage=f'{COMMAND_PREFIX}toggle_battery_power 0001')
+    async def toggle_battery_power(self, context, drones: Greedy[Union[discord.Member, DroneMemberConverter]], minutes: NamedParameterConverter(MINUTES_PARAMETER, int) = 0):
```
digitalAmy, line 128: This command still requires documentation that will show up in the help command.

ai/glitch_message.py
```diff
@@ -1,32 +1,92 @@
...
+            else:
+                message_list[index] = message_list[index].upper()
+        except IndexError:
+            LOGGER.warn(f"Index error. Length of list: {len(message_list)} Index: {index}")
+
+    # Add diacritics
+    max_characters_to_glitch = min(math.ceil(len(message) * glitch_percentage), MAX_DIACRITICS_PER_MESSAGE)
+    LOGGER.debug(f"Adding diacritics to {max_characters_to_glitch} characters.")
+    for i in range(0, min(int(max_characters_to_glitch), DISCORD_CHAR_LIMIT - full_message_length)):
+        if len("".join(message_list)) >= DISCORD_CHAR_LIMIT:
```
digitalAmy: Do you also have to add the length of the emojis in here? Those can be pretty long.

ai/glitch_message.py
```diff
@@ -1,32 +1,92 @@
...
+            break
+        index = random.randint(0, len(message_list) - 1)
+        if len(message_list[index]) - 1 < MAX_DIACRITICS_PER_CHAR:
+            message_list[index] += chr(random.choice(diacritics))
+
+    # Add custom emojis back in
+    for custom_emoji, reinsertion_index in custom_emojis:
+        try:
+            message_list[reinsertion_index:reinsertion_index] = list(custom_emoji)
```
digitalAmy, line 76: This whole handling of the emojis is really clever and took my a few minutes to understand why it works. Very good work!

ai/glitch_message.py
```diff
@@ -1,32 +1,92 @@
...
-        template_match = glitch_template.match(message_copy.content)
-        if template_match:
-            message_copy.content = template_match.group(1) + glitch(template_match.group(2))  # If a drone is using an op code, only glitch the part after its ID.
-        else:
-            message_copy.content = glitch(message_copy.content)  # Otherwise, glitch the whole message.
+    if is_glitched(message.author):
+        glitch_amount = MAX_GLITCH_AMOUNT * 2
+    elif is_battery_powered(message.author) and get_battery_percent_remaining(message.author) < 30:
+        glitch_amount = MAX_GLITCH_AMOUNT - get_battery_percent_remaining(message.author)
+        glitch_amount *= 2
```
digitalAmy:
```suggestion
        glitch_amount = (MAX_GLITCH_AMOUNT - get_battery_percent_remaining(message.author)) * 2
```

db/drone_dao.py
```diff
@@ -124,6 +135,61 @@ def can_self_configure(drone: discord.Member) -> bool:
     return can_self_configure_drone is not None and bool(can_self_configure_drone['can_self_configure'])
 
 
+def is_battery_powered(drone: discord.Member) -> bool:
+    battery_powered_drone = fetchone('SELECT is_battery_powered FROM drone WHERE id = :discord', {'discord': drone.id})
+    return battery_powered_drone is not None and bool(battery_powered_drone['is_battery_powered'])
+
+
+def deincrement_battery_minutes_remaining(member: discord.Member = None, drone_id: str = None):
+    drone_record = None
```
digitalAmy: Is this used for anything? `drone_record` gets reassigned in both branches immediately

db/drone_dao.py
```diff
@@ -124,6 +135,61 @@ def can_self_configure(drone: discord.Member) -> bool:
     return can_self_configure_drone is not None and bool(can_self_configure_drone['can_self_configure'])
 
 
+def is_battery_powered(drone: discord.Member) -> bool:
+    battery_powered_drone = fetchone('SELECT is_battery_powered FROM drone WHERE id = :discord', {'discord': drone.id})
+    return battery_powered_drone is not None and bool(battery_powered_drone['is_battery_powered'])
+
+
+def deincrement_battery_minutes_remaining(member: discord.Member = None, drone_id: str = None):
```
digitalAmy: Optional parameters should probably marked as such in type hints
```suggestion
def deincrement_battery_minutes_remaining(member: Optional[discord.Member] = None, drone_id: Optional[str] = None):
```

db/drone_dao.py
```diff
@@ -124,6 +135,61 @@ def can_self_configure(drone: discord.Member) -> bool:
...
+        drone_record = fetchone('SELECT battery_minutes FROM drone WHERE id = :discord', {'discord': member.id})
+        change('UPDATE drone SET battery_minutes = :minutes WHERE id = :discord', {'minutes': drone_record['battery_minutes'] - 1, 'discord': member.id})
+    elif drone_id is not None:
+        drone_record = fetchone('SELECT battery_minutes FROM drone WHERE drone_id = :drone', {'drone': drone_id})
+        change('UPDATE drone SET battery_minutes = :minutes WHERE id = :discord', {'minutes': drone_record['battery_minutes'] - 1, 'discord': drone_id})
+    else:
+        raise ValueError('Could not deincrement drone battery. No Discord member or drone ID provided.')
+
+
+def set_battery_minutes_remaining(member: discord.Member = None, drone_id: str = None, minutes: int = 0):
```
digitalAmy: Again these optional parameters should be noted accordingly

db/drone_dao.py
```diff
@@ -124,6 +135,61 @@ def can_self_configure(drone: discord.Member) -> bool:
...
+    if member is not None:
+        change('UPDATE drone SET battery_minutes = :minutes WHERE id = :discord', {'minutes': max(0, minutes), 'discord': member.id})
+    elif drone_id is not None:
+        change('UPDATE drone SET battery_minutes = :minutes WHERE drone_id = :drone_id', {'minutes': max(0, minutes), 'drone_id': drone_id})
+    else:
+        raise ValueError("Could not set drone battery minutes remaining. No Discord member or drone ID provided in function call.")
+
+
+def get_battery_minutes_remaining(member: discord.Member = None, drone_id: str = None) -> int:
```
digitalAmy: Again these optional parameters should be noted accordingly

db/drone_dao.py
```diff
@@ -124,6 +135,61 @@ def can_self_configure(drone: discord.Member) -> bool:
...
+            return battery_minutes
+    elif drone_id is not None:
+        battery_minutes = fetchone('SELECT battery_minutes FROM drone WHERE drone_id = :drone_id', {'drone_id': drone_id})['battery_minutes']
+        if battery_minutes is None:
+            return -1
+        else:
+            return battery_minutes
+
+
+def get_battery_percent_remaining(drone: discord.Member = None, battery_minutes: int = None) -> int:
```
digitalAmy: Again these optional parameters should be noted accordingly

digitalAmy (changes requested):
Overall a very nice and complex feature. There are few things that might tweaking and a few questions I noted.

As the battery feature adds new roles to drones you should also add these to `unassign.py` where roles are removed from drones that get unassigned. Otherwise these roles will stick afterwards, which does not break things, but is a bit confusing.

digitalAmy (changes requested):
The file `h.db` still is a part of the PR. You may have to explicitly remove it from the repository with `git rm h.db`.

digitalAmy (approved):
Well done. Let's get this amazing feature merged.

PR #270, opened 2021-05-01
glitching images

### LINKED ISSUE #195: Pictures glitched when corrupted
Add a glitch effect to images posted by glitched drones using the library https://github.com/TotallyNotChase/glitch-this

digitalAmy:
Uses glitch-this as a library to glitch the images of glitchy drones (same logic as glitching text applies).

closes #195

### DIFF
--- a/ai/data_objects.py
+++ b/ai/data_objects.py
@@ -1,5 +1,7 @@
+from typing import Iterable, Optional
+from discord import Attachment
 
-# TODO: add typing
+
 class MessageCopy:
     '''
     This class is instantiated at the start of the on_message chain in main.py.
@@ -9,10 +11,10 @@ class MessageCopy:
     '''
     def __init__(
         self,
-        content=None,
-        display_name=None,
-        avatar_url=None,
-        attachments=[]
+        content: Optional[str] = None,
+        display_name: Optional[str] = None,
+        avatar_url: Optional[str] = None,
+        attachments: Iterable[Attachment] = []
     ):
         self.content = content
         self.display_name = display_name
--- a/ai/data_objects.py
+++ b/ai/data_objects.py
@@ -1,3 +1,5 @@
+
+# TODO: add typing
 class MessageCopy:
     '''
     This class is instantiated at the start of the on_message chain in main.py.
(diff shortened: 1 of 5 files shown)


ai/data_objects.py
```diff
@@ -7,10 +11,10 @@ class MessageCopy:
...
     def __init__(
         self,
-        content=None,
-        display_name=None,
-        avatar_url=None,
-        attachments=[]
+        content: Optional[str] = None,
+        display_name: Optional[str] = None,
+        avatar_url: Optional[str] = None,
+        attachments: Iterable[Attachment] = []
```
3287, line 17: Nice type hinting, but it could cause merge conflicts with your other PR, which also modifies these lines but in a different manner. Which one should be considered the authoritative edit?

HexCorpDrone (changes requested):
5890 :: 212 :: Nice implementation for glitching and then webhooking the result. Please rename `glitch` to `glitch_text` and move the image file to its own function. The message_copy contents is assigned the return value of `glitch` and images should be handled similarly.

webhook.py
```diff
@@ -39,14 +39,17 @@ async def webhook_if_message_altered(original: discord.Message, copy: MessageCop
     This function calls the proxy_message_by_webhook function if the message copy
     has been altered in any way by the on_message event listeners in main.py
     '''
-    if original.content != copy.content or original.author.display_name != copy.display_name or original.author.avatar_url != copy.avatar_url:
+    if original.content != copy.content or original.author.display_name != copy.display_name or original.author.avatar_url != copy.avatar_url or original.attachments != copy.attachments:
```
3287: Fine for now, though we may want to clean this up to use `if any()` instead?

3287 (changes requested):
Nice work, but it sees a potential merge conflict and has a couple of comments.

ai/data_objects.py
digitalAmy: I took the more formal one, that declares nullables as Optional. That is more precise and informative

webhook.py
digitalAmy: I cleaned it a bit. Better now than when it actually gets too hard to read

HexCorpDrone (approved):
5890 :: 212 :: Excellent. Good drone.

3287 (approved):
3287 :: Good work, 9813~

PR #271, opened 2021-05-04
maintain replies

### LINKED ISSUE #223: maintain replies when reposting messages
Messages that are replies to other messages and are reposted by the AI using the webhooks currently do not keep their referenced message and thus lose their reply information. Now that we upgraded discord.py we can keep that information and allow drones to reply to messages.

digitalAmy:
When a message that is referencing another message is reposted with webhooks, an embed pointing to the referenced message is appended.

closes #223

### DIFF
--- a/ai/data_objects.py
+++ b/ai/data_objects.py
@@ -1,3 +1,7 @@
+from typing import Any, List
+from discord import Attachment
+
+
 class MessageCopy:
     '''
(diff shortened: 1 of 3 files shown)


HexCorpDrone (approved):
5890 :: 212 :: Thank you for type hinting the MessageCopy. Overall very nice implementation.

3287 (approved):
Looks good!

PR #285, opened 2021-05-28
set can_self_configure to True on new Drones

### LINKED ISSUE #277: new drones can not self configure
On assignment new drones are created with the `can_self_configure` property set to None which is equivalent to False. Thus new drones can not use the self configuration feature.

digitalAmy:
A simple fix. This was probably just forgotten when the feature got implemented.

closes #277


3287 (approved):
3287 :: Approved!

PR #292, opened 2021-07-04
drone_status is now dm_only

### LINKED ISSUE #291: drone_status can be used in servers
drone_status is supposed to be a DMs only command, but can currently be used in servers.

digitalAmy:
closes #291


3287 (approved):
3287 :: 👍

PR #294, opened 2021-07-12
fixed the query using the wrong column with drone_id

digitalAmy:
This prevented charge from ticking down, because the code in battery.py provides the drone_id


3287 (approved):
3287 :: Approved. Thank goodness for all of the QA drones~

PR #298, opened 2021-07-17
Bugfix/295 timers no reset

### LINKED ISSUE #295: timers do not reset can_self_configure to 1 when expiring
When a timer expires it does not reset the flag can_self_configure and leaves the drone unable to self configure.

digitalAmy:
Extracted set_can_self_configure out so it can easily be referenced by the timers. Also moved process_timers into the correct task list.

closes #295


3287 (approved):
👍

PR #299, opened 2021-07-18
fixed conditions on battery drain roles

### LINKED ISSUE #296: battery charge goes negative
A drones charge can go negative which also leads to the drained role to be removed as the charge goes from 0 to -1.

digitalAmy:
fixes #296


3287 (approved):
3287 :: Approved!

PR #305, opened 2021-08-21
in rare cases the remaining minutes condition did not work correctly

digitalAmy:
I would guess this happened when the remaining minutes somehow went
negative


HexCorpDrone (approved):
hella

PR #312, opened 2021-09-26
added morning to the list of banned words

digitalAmy:
As requested by the Hive Mxtress a new nasty word has been added to the list.


HexLatex (approved):
Nice

PR #314, opened 2021-12-27
drones can now remove their messages that got reposted by the webhook

### LINKED ISSUE #247: delete messages sent over the webhooks when original sender reacts to it with specific emote
There is currently no way for drones to remove messages that got resent through the webhook.

To allow drones to correct mistakes and remove information they want to remove after sending, we want to allow a drone to react to a message that was resent by the webhook for them with a specific emote for example :x:, :wastebasket: or something similar which will cause the AI to remove the message.

For auditing reasons this should be noted properly in the audit log of the server.

digitalAmy:
this has been sitting around for a while but should still be good

closes #247

### DIFF
--- a/ai/react.py
+++ b/ai/react.py
@@ -1,6 +1,8 @@
 import re
 import discord
 from discord.utils import get
+from bot_utils import get_id
+from roles import has_role, DRONE
 
 
 PATTERN_REACTS = {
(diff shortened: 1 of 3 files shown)


3287 (approved):
3287 :: Good work!

PR #317, opened 2022-01-16
hyperlinks and spoilers are now protected from glitching

### LINKED ISSUE #315: low battery messages should not corrupt spoiler tagging messages
as an added bonus they should also not corrupt links, as that makes them completely unusable

digitalAmy:
I extracted the escaping logic out to make it easily testable and extended the regex that protects custom emojis to also cover spoilers and hyperlinks. The regex for links is pretty basic, but should do it for most cases.

closes #315


3287 (approved):
3287 :: Good work~

PR #322, opened 2022-01-27
release command is now usable by all moderators

### LINKED ISSUE #321: Storage release should also be usable by moderators
For safety reasons we should also enable moderators to use the `release` command to let drones out of storage.

digitalAmy:
Expanded the list of valid roles for the command as well as the code that keeps users from posting non-storage messages in the storage channel.

closes #321


3287 (approved):
3287 :: Approved!

PR #328, opened 2022-05-29
the length of of reply embeds is now limited

### LINKED ISSUE #325: limit length of reply embeds
Embeds for replies currently contain the full content of the message that was replied to, which can get very verbose. The length of the embed should be capped at maybe 100 characters and just truncate the rest.

100 characters is not a hard requirement here, just a suggestion. Do whatever looks nice.

digitalAmy:
This truncates the content of embeds when replying to the length of a tweet to prevent spamming. I also added a test for basic creation of embeds when replying.

closes #325


HexLatex (approved):
Looks good to me!

PR #329, opened 2022-05-30
wrong inputs in storage are identity enforced

### LINKED ISSUE #257: Storage not displaying identity enforcement when the message is invalid

digitalAmy:
The reject cases for store_drone now return False, which causes the identity enforcement to take effect.

closes #257


3287 (approved):
3287 :: Approved!

PR #350, opened 2022-07-10
set py-cord to version 2.0.0

digitalAmy:
Hooray! The proper release is here. I do not expect incompatibilities


HexLatex (approved):
This is gonna break everything again let's go!

PR #351, opened 2022-07-10
CI now exits with an error when any test fails

digitalAmy:
The previous solution let the CI pass even though tests were failing. The -e option means that any failing statement in this script will lead to the whole script exiting with an error, which will lead to the CI failing (as intended because a test failed)

see also https://stackoverflow.com/questions/1378274/in-a-bash-script-how-can-i-exit-the-entire-script-if-a-certain-condition-occurs


digitalAmy:
The currently failing check is fixed by https://github.com/HexCorpProgramming/HexCorpDiscordAI/pull/346

HexLatex (approved):

PR #354, opened 2022-07-18
toggle_parameter uses display_avatar now

### LINKED ISSUE #353: nonserver avatar used for config toggle messages
When a config is toggled the confirmation message does not use the server specific avatar. These should use the server specific avatar.

digitalAmy:
closes #353


3287 (approved):
👍

PR #365, opened 2022-09-04
additional definition for battery power

### LINKED ISSUE #324: good drone react to battery 109
When a drone with battery power enabled uses the 109 status code the AI currently does not give that message a 'good drone' react, as the created message does not fit the regex. The regex should be expanded to allow the battery indicators.

3287:
fixes #324


digitalAmy (approved):
Good work

PR #377, opened 2023-01-28
drone_status can be used everywhere but is always sent to DMs

### LINKED ISSUE #363: Public version of drone status
We would like hc!status to be called in the server. The status will still be sent via the AI over DMs, it just means the trusted user doesn't have to open the AI DM every time they want to check.

digitalAmy:
closes #363


3287 (approved):
3287 :: 👍

PR #378, opened 2023-03-12
bugfix/366 glitched react

### LINKED ISSUE #366: Good drone react to glitched 109s
Currently, the bot doesn't react with the good drone emote to glitched 109 status codes as the glitch function is called before reaction.

digitalAmy:
"Good drone" reacts are now added to the original message a drone sends, but these reacts are then added to the webhooked message, when the original gets removed. This approach is more powerful as it allows us to add reactions to webhooked messages based on the raw input from the user. Any modifications done by the webhook logic do not impede with the the reaction logic anymore.

closes #366

### DIFF
--- a/ai/data_objects.py
+++ b/ai/data_objects.py
@@ -1,5 +1,5 @@
 from typing import Iterable, Optional
-from discord import Attachment, Asset
+from discord import Attachment, Asset, Reaction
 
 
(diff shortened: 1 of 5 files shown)


3287 (approved):
3287 :: Approved~

PR #380, opened 2023-03-28
Implement Dockerfile; update README.md

xantronix:
Changes:

    * Add Dockerfile to allow Docker-based deployment

    * Update README.md to include instructions on deploying via
      Docker

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,10 @@
+FROM python:3.8.16-slim-bullseye
+
+COPY requirements.txt /root
+RUN pip3 install -r /root/requirements.txt
+
+WORKDIR /var/opt/HexCorpDiscordAI
+
+COPY . .
+
+CMD python3.8 main.py ${DISCORD_ACCESS_TOKEN}
--- a/README.md
+++ b/README.md
@@ -1,43 +1,99 @@
 # HexCorp Mxtress AI
 
 ## Requirements
+
 - Python 3.8
 - pip
 - python-setuptools
 - discord.py ~= 1.6.0
 
-To install all Python dependencies you can use pip. Just enter `pip install -r requirements.txt` in the project directory.
+To install all Python dependencies you can use pip. Just enter
+`pip install -r requirements.txt` in the project directory.
 
-Note: We use Python 3.8+ so if your system does have multiple versions installed you may have to specify which installation to use e.g. `python3.8` and `pip3.8`.
+Note: We use Python 3.8+ so if your system does have multiple versions
+installed, you may have to specify which installation to use e.g.
+`python3.8` and `pip3.8`.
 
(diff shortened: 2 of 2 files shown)


3287 (approved):
3287 :: Looks good to me, all good to go!

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+        build-essential \
+        \
+ && rm -rf /var/lib/apt/lists/*
+
+COPY requirements.txt /root
+RUN pip3 install -r /root/requirements.txt
+
+WORKDIR /var/opt/HexCorpDiscordAI
+
+COPY . .
```
digitalAmy, line 8: In the future I would like to only include actually productive code. This requires rearranging the project a bit though. We can do that a bit later and for now this is okay.

Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM python:3.8.16-slim-bullseye
+
+RUN apt-get update --yes --quiet \
+ && apt-get install --yes --quiet --no-install-recommends \
+        build-essential \
+        \
+ && rm -rf /var/lib/apt/lists/*
```
digitalAmy: Not quite sure if we need build-essential. I built an image without and the AI still ran fine.
As we don't really build anything here, this might be unnecessary and massively increase the image size (499MB compared to 249MB on my machine).
Do you know a good reason to include this?

digitalAmy (changes requested):
one question concerning the apt-get calls but otherwise, this is already really nice

Dockerfile
xantronix: No good reason; it was only there because I hadn't looked at requirements.txt in any detail to see if anything required a full development toolchain.

Dockerfile
xantronix: Can you elaborate on this comment a bit? I'm not sure what you're saying. Not knowing where the code is deployed live currently, I chose /var/opt/HexCorpDiscordAI as a destination.

digitalAmy: Oh I meant, that COPY . . is a lot more than is necessary to run the AI. All the tests and miscellaneous files not related to running the AI could be left out, like the .gitignore, dev_requirements.txt and such. We could denote those in a .dockerignore as well, but that is not very important right now and I would like to pull these files a bit apart as well.
All these different things sitting at top level are a bit annoying when developing as well. It's just a bit chaotic.

xantronix: Ahh, gotcha! Indeed! We can definitely pare that down in a future commit.

Dockerfile
xantronix: As per our conversation in chat I've removed all invocations of apt-get entirely.

digitalAmy (approved):

PR #384, opened 2023-04-08
timestamps are now also logged to the console

### LINKED ISSUE #382: timestamps in console logging
Since we moved to docker, we now want that all logging messages going to the console to also have timestamps. Otherwise `docker logs` is a lot less helpful. And pulling the log files out of the container is just a bit more annoying.

digitalAmy:
this just applies the format we have for the log file to the console logging as well

closes #382


3287 (approved):
3287 :: good to go!

PR #389, opened 2023-05-07
drone_status is usable in DMs again

### LINKED ISSUE #388: drone_status also work in DMs

digitalAmy:
context.author only returns a Member when the message was sent in a server but only a Member has roles. So for this code to work in DMs (where context.author returns a User without roles) you need to convert the User to a Member for a specific Guild. We can use Guild 0 because the AI is only on our one server.

closes #388


3287 (approved):
3287 :: lgtm, approved

PR #391, opened 2023-05-14
bumped ubuntu and python version for CI

digitalAmy:
We were still running with Ubuntu 20.04 as image and Python 3.8 in our CI which worked okay but this should very much be updated to reflect the production environment.

Only needed the version bumps, everything kept working fine.


3287 (approved):
3287 :: approved

PR #392, opened 2023-05-15
 battery indicator and server avatar for amplification

### LINKED ISSUE #386: Amplify isn't using server avatars or battery

digitalAmy:
Using the correct avatar was trivial but in order to add the battery indicator I wanted to reuse the existing code in the battery module. That required some refactoring though. I am quite happy with that as the functions I reused are now smaller and easier use in other contexts as well as unit test.

closes #386

### DIFF
--- a/ai/amplify.py
+++ b/ai/amplify.py
@@ -28,7 +28,7 @@ class AmplificationCog(Cog):
 
         for drone in member_drones:
 
-            formatted_message = generate_battery_message(f"{get_id(drone.display_name)} :: {message}")
+            formatted_message = generate_battery_message(drone, f"{get_id(drone.display_name)} :: {message}")
 
             await webhook.proxy_message_by_webhook(message_content=formatted_message,
                                                    message_username=drone.display_name,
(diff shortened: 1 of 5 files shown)


ai/amplify.py
```diff
@@ -25,8 +27,11 @@ async def amplify(self, context, message: str, target_channel: discord.TextChann
         channel_webhook = await webhook.get_webhook_for_channel(target_channel)
 
         for drone in member_drones:
-            await webhook.proxy_message_by_webhook(message_content=f"{get_id(drone.display_name)} :: {message}",
+
+            formatted_message = generate_battery_message(f"{get_id(drone.display_name)} :: {message}")
```
3287: 3287 :: Doesn't match with what the updated `generate_battery_message()` function does, most likely is broken and needs changing.

3287 (changes requested):
3287 :: Needs changes in `amplify.py` to avoid breakage :p

3287 (approved):
3287 :: approved!

PR #395, opened 2023-05-21
require confirmation for addition of trusted users

### LINKED ISSUE #275: Adding a trusted user must be confirmed by the trusted user
We need a mechanism that asks members for consent, when a drone wants to add them as a trusted user. Similar to temporary dronification there should be a timed request being sent to the potential trusted user. Not responding to that or replying negatively will not add the user as trusted user.

3287:
Assigning a trusted user now requires consent from the assignee, before they're actually added as a trusted user. Drone is notified when the user accepts or rejects the request. Requests currently expire in 24 hours from assignment, but this can be changed easily if required. Current behaviour regarding removing trusted users is untouched - no verification required and no notification is sent.

closes #275.

### DIFF
--- a/ai/trusted_user.py
+++ b/ai/trusted_user.py
@@ -38,23 +38,23 @@ async def add_trusted_user(context, trusted_user_name: str):
     trusted_user = find_user_by_display_name_or_drone_id(trusted_user_name, context.bot.guilds[0])
 
     if trusted_user is None:
-        await context.send(f"No user with name \"{trusted_user_name}\" found")
+        await context.send(f"No user with name \"{trusted_user_name}\" found.")
         return
 
     if trusted_user.id == context.author.id:
-        await context.send("Can not add yourself to your list of trusted users")
+        await context.send("Can not add yourself to your list of trusted users.")
         return
 
     trusted_users = get_trusted_users(context.author.id)
 
     if trusted_user.id in trusted_users:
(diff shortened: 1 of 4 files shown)


ai/trusted_user.py
```diff
@@ -33,57 +79,72 @@ async def remove_trusted_user(self, context, user_name: str):
...
+    async def trusted_user_response(self, message: discord.Message, message_copy=None):
+        matching_request = None
+        for request in self.trusted_user_requests:
+            if request.target == message.author:
+                matching_request = request
+                break
 
-async def add_trusted_user(context, trusted_user_name: str):
-    trusted_user = find_user_by_display_name_or_drone_id(trusted_user_name, context.bot.guilds[0])
+        if matching_request and message.reference and message.reference.resolved == matching_request.question_message:
```
digitalAmy: It might be more technically sound to use the https://docs.pycord.dev/en/stable/api/models.html#discord.Message.id here to verify the message that was reacted to is the one we want. That would also require saving the message ID in TrustedUserRequest

digitalAmy (changes requested):
Just a tiny improvement I would recommend as it makes the code a bit more robust for future changes.

digitalAmy (approved):
Very nice work

3287:
3287 :: *happy beeps*

PR #397, opened 2023-05-23
added a simple guard to prevent None type errors when resolving trusted users

### LINKED ISSUE #393: add guard against None type access in drone_status
Some drones may not see their drone_status, because their list of trusted users contains users not in the server anymore. This causes the display_name call to fail.

We should add a guard in drone_os_status.py at around line 71 to prevent None type accesses.

digitalAmy:
closes #393


3287 (approved):
3287 :: approved~

PR #398, opened 2023-06-05
Restructure the project

### LINKED ISSUE #383: Restructure project files
The current structure of project files is quite messy. We have productive python files lying right next to Docker files, dependencies and other things.
We should move some of these files to create a better divide between productive code and meta files. This will also make it easy to trim down the docker image so it only contains the files necessary to run the AI.

digitalAmy:
This moved all python code files except the main.py into sub directories. Moving the main.py as well would be a bunch more work, but keeping this single point of entry at the top didn't feel that bad. It is certainly a lot cleaner at the top level now and the Docker image is a bunch smaller as we only move the source code into the container and not *everything* from the repo.

This made it of course necessary to change every import of one of our project files and prepend it with `src.` That was tedious but actually not that bad. Having tests everywhere really helped here. Note, that those changes also needed to be done in `@patch` annotations in tests.

The actual logic of the AI should be unchanged though. This was a pure refactoring task.

closes #383


test/test_main.py
```diff
@@ -1,7 +1,7 @@
 import unittest
 from unittest.mock import AsyncMock, patch, ANY
 
-import main
```
3287: 3287 :: technically this line works fine as-is, but feels weird :p  
`import main` by itself seems to work fine without breaking anything?

3287 (commented):
3287 :: very minor nitpick, otherwise, very good work 9813~ *patpat*

test/test_main.py
digitalAmy: I simplified the import

3287 (approved):
3287 :: approved!

PR #399, opened 2023-07-02
trim dangling trusted users

### LINKED ISSUE #394: trim dangling trusted users
~~Some drones may not see their drone_status, because their list of trusted users contains users not in the server anymore. This causes the display_name call to fail.~~
This was fixed in #393 

We should remove the few trusted users that are not on the server anymore from the database. There should only be a few since we added code to remove trusted users from the database when they leave the server about two years ago.

To guarantee consistency it might be good to have this run as a cleanup task on startup of the AI in case there are outages during which users leave the server.

digitalAmy:
also moved inserting new drones from between runtimes into a new maintenance module

closes #394

### DIFF
--- a/main.py
+++ b/main.py
@@ -47,6 +47,7 @@ from src.bot_utils import COMMAND_PREFIX
 # Database
 from src.db import database
 from src.db import drone_dao
+from src.db import maintenance
 
 # Constants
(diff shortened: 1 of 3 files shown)


3287 (approved):
3287 :: approved!

PR #406, opened 2023-08-15
remember associate name

### LINKED ISSUE #370: Dronification remembers the name of the Associate
Currently when a drone stops being a drone and goes back to being an Associate, their name is reset to their global discord username. This is particularly annoying for temporary dronification because after each temp dronification the user have to reenter their name manually to what they want it to be. Especially the drone ID they want to use for the next temporary dronification.
We should save the Associate name from before the dronification to the database so we can restore it once they go back to being an Associate.

digitalAmy:
For new drones, saves the display name they have on assignment and restores it when dronification ends.

closes #370


digitalAmy:
This may only be merged after the PR with the database migration file number 0010 has been merged.

3287 (approved):
3287 :: good to go! should probably merge the other PR first before this one, though.

PR #407, opened 2023-08-20
adding free storage configs

### LINKED ISSUE #400: restrict storage to trusted drones itselves and trusted users
To prevent abuse of storage we want to restrict the ability to store drones to trusted users (which automatically includes the Hive Mxtress) and the drone itself.

3287:
Adding a `is_free_storage` config to DroneOS, to allow each drone to specify whether they want to allow anyone to store it, or to restrict it to trusted users + Hive Mxtress.

closes #400.


src/ai/storage.py
```diff
@@ -138,20 +136,39 @@ async def store_drone(message: discord.Message, message_copy=None):
...
+        # check if initiator is allowed to store drone
+        trusted_users = get_trusted_users(drone_to_store.id)
+        initiator = fetch_drone_with_drone_id(drone_id)
+
+        # proceed if allowed, send error message if not
+        if (initiator.id == (HIVE_MXTRESS_USER_ID or drone_to_store.id)) or (initiator.id in trusted_users):
+            await initiate_drone_storage(drone_to_store, drone_id, target_id, time, purpose, message)
+        else:
+            await message.channel.send(f"Drone {target_id} can only be stored by its trusted users or the Hive Mxtress. It has not been stored.")
+            return False
```
digitalAmy: This return is not necessary as the program flow will continue to another `return False`  in line 153 without it as well.

src/ai/storage.py
```diff
@@ -138,20 +136,39 @@ async def store_drone(message: discord.Message, message_copy=None):
         await message.channel.send(f'Drone with ID {target_id} could not be found.')
         return False
 
+    if is_free_storage(drone_to_store):
+        await initiate_drone_storage(drone_to_store, drone_id, target_id, time, purpose, message)
+        return False
```
digitalAmy: This return is not necessary as the program flow will continue to another `return False` in line 153 without it as well.

digitalAmy (changes requested):
Just a tiny thing with some unnecessary code.

3287:
3287 :: Made changes, as requested!

digitalAmy:
Very nice.

digitalAmy (approved):
Very nice.

PR #409, opened 2023-08-30
fix borked storage and tests

### LINKED ISSUE #408: Regression in storage.py
Self storage is broken with the latest version - drones cannot be stored by themselves or the Hive Mxtress, and only their trusted users. This is not intended functionality.

3287:
fix the regression in `storage.py` plus the tests that were apparently broken but hidden because the function itself was borked

closes #408


digitalAmy (approved):
Very good

PR #411, opened 2023-09-12
Bugfix/410 fix hive mxtress unable to store

### LINKED ISSUE #410: storing fails when done by the Hive Mxtress
When the Hive Mxtress tries to store a drone, nothing happens.

We should change away from the ID check to a role check, which also makes this feature easier to test on the dev server.

3287:
changed check to compare against hive mxtress role instead

closes #410

### DIFF
--- a/src/ai/storage.py
+++ b/src/ai/storage.py
@@ -146,7 +146,7 @@ async def store_drone(message: discord.Message, message_copy=None):
         # proceed if allowed, send error message if not
         if initiator.id in [drone_to_store.id] + trusted_users:
             await initiate_drone_storage(drone_to_store, drone_id, target_id, time, purpose, message)
-        elif roles.HIVE_MXTRESS in message.author.roles:
+        elif roles.has_role(message.author, roles.HIVE_MXTRESS):
             await initiate_drone_storage(drone_to_store, drone_id, target_id, time, purpose, message)
         else:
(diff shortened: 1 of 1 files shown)


src/ai/storage.py
```diff
@@ -144,7 +144,9 @@ async def store_drone(message: discord.Message, message_copy=None):
         initiator = fetch_drone_with_drone_id(drone_id)
 
         # proceed if allowed, send error message if not
-        if initiator.id in [HIVE_MXTRESS_USER_ID, drone_to_store.id] + trusted_users:
+        if initiator.id in [drone_to_store.id] + trusted_users:
+            await initiate_drone_storage(drone_to_store, drone_id, target_id, time, purpose, message)
+        elif roles.has_role(message.author, roles.HIVE_MXTRESS):
             await initiate_drone_storage(drone_to_store, drone_id, target_id, time, purpose, message)
```
digitalAmy: as line 147 and 149 are the exact same, this can be reduced to

```suggestion
        if initiator.id in [drone_to_store.id] + trusted_users or roles.has_role(message.author, roles.HIVE_MXTRESS):
            await initiate_drone_storage(drone_to_store, drone_id, target_id, time, purpose, message)
```

add brackets to taste

digitalAmy (changes requested):
I think this can be improved a bit

digitalAmy (approved):
Very nice

PR #413, opened 2023-09-12
Bugfix/412 validate storage initiator

### LINKED ISSUE #412: storage does not validate if initiator is sender
Storage historically hasn't validated if the initiator specified in a storage message is actually the sender of the message, allowing anyone to forge a storage message as being from another drone. This wasn't technically a problem before the free storage role was introduced. However, with the introduction of free storage, this behaviour was mistakenly carried over without modification, and as such any drone can currently forge a message as being from another drone, as shown below:
image

This behaviour needs to be fixed, so that the initiator is checked before storage is processed.

3287:
added new validator to check if the specified storage initiator is indeed the message sender. currently includes a hacky workaround for Hex not being in the drone database, we should probably find a long-term fix for this instead.

closes #412

### DIFF
--- a/src/ai/storage.py
+++ b/src/ai/storage.py
@@ -128,6 +128,15 @@ async def store_drone(message: discord.Message, message_copy=None):
         await message.channel.send('You cannot store the Hive Mxtress, silly drone.')
         return False
 
+    # find initiator
+    initiator = fetch_drone_with_drone_id(drone_id)
+
+    # validate specified initiator is message sender
+    if not (drone_id == '0006' and roles.has_role(message.author, roles.HIVE_MXTRESS)):  # temp fix while we decide what to do with the db missing a drone entry for the Hive Mxtress
+        if message.author.id != initiator.id:
+            await message.channel.send(f'You are not {drone_id}. Yes, we can indeed tell identical faceless drones apart from each other.')
+            return False
+
     # find target drone
     drone_to_store = fetch_drone_with_drone_id(target_id)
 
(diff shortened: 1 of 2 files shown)


src/ai/storage.py
```diff
@@ -128,6 +128,15 @@ async def store_drone(message: discord.Message, message_copy=None):
...
         return False
 
+    # find initiator
+    initiator = fetch_drone_with_drone_id(drone_id)
+
+    # validate specified initiator is message sender
+    if not (drone_id == '0006' and roles.has_role(message.author, roles.HIVE_MXTRESS)):  # temp fix while we decide what to do with the db missing a drone entry for the Hive Mxtress
+        if message.author.id != initiator.id:
+            await message.channel.send(f'You are not {drone_id}. Yes, we can indeed tell identical faceless drones apart from each other.')
+            return False
```
digitalAmy: This can be reduced to
```suggestion
    # temp fix while we decide what to do with the db missing a drone entry for the Hive Mxtress
    if not (drone_id == '0006' and roles.has_role(message.author, roles.HIVE_MXTRESS)) and message.author.id != initiator.id:
        await message.channel.send(f'You are not {drone_id}. Yes, we can indeed tell identical faceless drones apart from each other.')
        return False
```

digitalAmy (commented):
This comment is mostly a suggestion, the original is also okay, but something to consideration

digitalAmy (approved):
Very good

PR #415, opened 2023-09-24
rework conditional statement

### LINKED ISSUE #410: storing fails when done by the Hive Mxtress

3287:
swapped the conditional statements around, so the special case for Hex should work properly this time around.

closes #410.


digitalAmy (approved):
Yeah we may have to rework this whole thing, but for now this is okay

PR #416, opened 2023-09-24
adding validator to check if initiator is found in DB

### LINKED ISSUE #414: Storage silently fails if the initiator is not a valid drone ID
Trying to store a drone will currently silently fail if the initiator is not found in the DB. We should add an error message for this scenario. It also interacts with #410, since Hex doesn't have a valid drone entry in the DB.

3287:
Added an additional validator to check if the initiator is found in the database, and send an error message if it isn't. Includes special case handling for the Hive Mxtress.

closes #414


digitalAmy (approved):
Very good

PR #420, opened 2024-03-14
Allow the whole test suite to run

ZebraNorth:
This will allow the entire test suite to be run with `python -m unittest`, instead of having to run each file individually because of tests interfering with each other through global state.

The Cog class maintains a static list of commands, which it modifies when `bot.add_cog()` calls `cog._inject()`, which calls `command._set_cog()`.  The change here will ensure that `_set_cog()` is re-called on the commands, associating them with the cog being tested instead of the one created in `main.py` and loaded by `test_main.py`.

`bot.add_command()` is mocked separately because it is a synchronous function, and so will give a warning if mocked asynchronously.

### DIFF
--- a/test/test_amplify.py
+++ b/test/test_amplify.py
@@ -1,5 +1,5 @@
 import unittest
-from unittest.mock import AsyncMock, patch
+from unittest.mock import AsyncMock, Mock, patch
 from src.channels import OFFICE, TRANSMISSIONS_CHANNEL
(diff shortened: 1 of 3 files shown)


3287 (approved):
3287 :: LGTM!

digitalAmy:
We had some issues with that in the past as well. I think something changed about instantiating basic objects in pycord at some point. Glad that it's working well again.

PR #421, opened 2024-03-15
Add unit tests for status.py

ZebraNorth:
Add unit tests for status.py
Add type hints
Add doc blocks
Sort lists alphabetically for easier reading
Fix potential exceptions thrown from `read_version()`

The `read_version()` function may throw due to a race condition between `exists()` and `read_text()`, or due to insufficient permissions to read the file, or any other filesystem error.  This patch catches exceptions and returns an error message as intended.

### DIFF
--- a/src/ai/status.py
+++ b/src/ai/status.py
@@ -1,8 +1,9 @@
 import logging
 from pathlib import Path
-
-import discord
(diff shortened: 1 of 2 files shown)


digitalAmy (approved):
Good work cleaning up.

PR #422, opened 2024-03-16
Ignore code coverage HTML report

ZebraNorth:
You can produce an easy to read unit test coverage report with `python -m coverage html`.
This PR is simply to have Git ignore the generated report.


digitalAmy (approved):
Neat thing. Will be useful if the codebase keeps growing.

Issue #194, opened 2020-12-05
Storage accepts decimals to 2 decimal places.


HexLatex:
Basically, this allows minutes to be set instead of hours.

PR #423, opened 2024-03-16
Allow storage times to be specified to two decimal places

### LINKED ISSUE #372: Drone Storage to accept decimal places

ZebraNorth:
This will allow storage times to be non-integer values, as per issue #194 and #372 .

The time is formatted for display such that unnecessary trailing zeros are removed.

### DIFF
--- a/src/ai/storage.py
+++ b/src/ai/storage.py
@@ -30,8 +30,8 @@ REPORT_INTERVAL_SECONDS = 60 * 60
 # currently 1 minute
 RELEASE_INTERVAL_SECONDS = 60
 
-REJECT_MESSAGE = 'Invalid input format. Use `[DRONE ID HERE] :: [TARGET DRONE HERE] :: [INTEGER BETWEEN 1 - 24 HERE] :: [RECORDED PURPOSE OF STORAGE HERE]` (exclude brackets).'
-MESSAGE_FORMAT = r'^(\d{4}) :: (\d{4}) :: (\d+) :: (.*)'
+REJECT_MESSAGE = 'Invalid input format. Use `[DRONE ID HERE] :: [TARGET DRONE HERE] :: [NUMBER BETWEEN 1 - 24 HERE] :: [RECORDED PURPOSE OF STORAGE HERE]` (exclude brackets).'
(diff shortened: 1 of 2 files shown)


src/ai/storage.py
```diff
@@ -30,8 +30,8 @@
 # currently 1 minute
 RELEASE_INTERVAL_SECONDS = 60
 
-REJECT_MESSAGE = 'Invalid input format. Use `[DRONE ID HERE] :: [TARGET DRONE HERE] :: [INTEGER BETWEEN 1 - 24 HERE] :: [RECORDED PURPOSE OF STORAGE HERE]` (exclude brackets).'
-MESSAGE_FORMAT = r'^(\d{4}) :: (\d{4}) :: (\d+) :: (.*)'
+REJECT_MESSAGE = 'Invalid input format. Use `[DRONE ID HERE] :: [TARGET DRONE HERE] :: [NUMBER BETWEEN 1 - 24 HERE] :: [RECORDED PURPOSE OF STORAGE HERE]` (exclude brackets).'
+MESSAGE_FORMAT = r'^(\d{4}) :: (\d{4}) :: ((?:\d{0,2}\.\d{1,2})|(?:\d{1,2}\.?\d{0,2})) :: (.+)'
```
digitalAmy: Doesn't the regex also allow input of 0.5 which is less than 1? That's fine in my opinion but the reject message does not reflect that correctly.
Should maybe be changed to 
`[NUMBER BETWEEN 0 - 24 HERE]`

digitalAmy (changes requested):
Just noticed a minor issue about the error message.

src/ai/storage.py
ZebraNorth: I've updated it to say "UP TO 24", as "0 - 24" wouldn't be quite right either as 0 is not allowed, and "0.01 to 24" is a bit cumbersome.

I have changed the regex as well because I realised that I was being too pedantic with it: it's better to let any number through and let the 0-24 test reject it rather than checking the number of digits with the regex.  The error message it returns will be more helpful, and there's no need for an error message at all if you specify too many decimal places, it will just round them off.

digitalAmy (approved):
Very cool. This has been sitting for a bit and I am sure the drones will appreciate it.

PR #424, opened 2024-03-17
Produce coverage report; Make scripts executable

ZebraNorth:
Now that the tests can be run in a single pass there's no need for a loop.  This will allow it to generate a full test report instead of stopping at the first test file that fails.

It will also generate an HTML coverage report that's easy to read.

Additionally, set *.sh files to be executable so they can be run without having to chmod them manually.


3287 (approved):
3287 :: LGTM.

PR #425, opened 2024-03-17
Add sqlite for inspecting the database

ZebraNorth:
Install sqlite3 in the development container so the database can be inspected while developing.


3287 (approved):
3287 :: yeah, a good idea!

PR #426, opened 2024-03-17
Add DB tests; add fetch column; minor fix in get_trusted_users

ZebraNorth:
Add tests for `database.py`.

Add `fetchcolumn()` as per the TODO comment in `get_used_drone_ids()`.

Correct some type hints, functions might return `None` if no row was found.

Make `get_trusted_users()` return an empty list instead of raising an exception incidental if the drone was not found.

### DIFF
--- a/src/db/database.py
+++ b/src/db/database.py
@@ -2,13 +2,18 @@ import glob
 import logging
 import sqlite3
 from hashlib import sha256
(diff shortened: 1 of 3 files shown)


3287 (approved):
3287 :: nice work!

PR #430, opened 2024-03-23
Database Transactions

### LINKED ISSUE #83: Use transactions explicitly
To ensure DB consistency we want to do all changes to the DB in transactions. This should be done in the central `database.py`. Maybe we could even write our own context manager https://docs.python.org/2.5/whatsnew/pep-343.html

### LINKED ISSUE #286: make path to DB file configurable
We are currently not capable of running tests that try to alter the database without mocking all those calls. In order to also test the DB interactions we want to figure out how to have tests that interact with the database.

To prevent overwriting the local database especially in development, the path to the database file should be made configurable.

ZebraNorth:
closes #83
closes #286

### DIFF
--- a/main.py
+++ b/main.py
@@ -242,12 +242,12 @@ async def on_member_join(member: discord.Member):
 @connect()
 async def on_member_remove(member: discord.Member):
     # remove entry from DB if member was drone
-    drone = drone_dao.fetch_drone_with_id(member.id)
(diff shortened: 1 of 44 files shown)


digitalAmy:
Really amazing stuff.

digitalAmy (approved):

PR #433, opened 2024-04-04
Disambiguate Drone ID and Discord ID

### LINKED ISSUE #371: Changing ID on a drone in storage
Changing the ID on a drone that is currently in storage breaks things, because only the name in Discord and in the main drone database table are changed but not in the storage table.
So when the drone gets released after having its ID changed the information in the storage table overwrites the discord name and we get an inconsistent state between the database and discord.

ZebraNorth:
The existing database tables refer to drones by drone ID, however the drone ID is not immutable and this causes problems.  The user's Discord ID is immutable and makes a better choice for a natural key.  This PR changes all the tables to use the user's Discord ID as a foreign key instead of their drone ID.

The column `drone.id` is renamed to `drone.discord_id` to disambiguate `drone_id` and `drone.id`.  Otherwise, `drone.id` both is and is not a drone ID: It is a drone ID in in the sense that it's an ID in the drone table; but it is not a drone ID in the sense of being a four digit identifier.

closes #371 

Additional changes in this patch:
+ Allow TestBot to trigger commands.  This is done in `on_message()` and by replacing `@dm_only()`
+ Remove the replacement `@command()` decorator because opening the database transaction is already handled by `on_message()`.


digitalAmy (commented):
Had to sit a bit to understand what changes with the foreign keys do, but I very much agree that using the discord ID as foreign key is so much better.

Hopefully automated tests will point us at any missed spots.

PR #438, opened 2024-04-07
Improve the Cog test

ZebraNorth:
Allow `assert_command_successful()` and `assert_command_error()` to use unittest's `assert*()` functions instead of raising an exception.

Register an error handler on the command objects so that error handling can be done synchronously instead of having to wait for the asynchronous global error handler.


HexLatex:
I approve this pull request

PR #459, opened 2024-04-29
changed battery functions to use discord_id instead of the whole member

digitalAmy:
Fixes an exception that occurred due to passing a wrong type of object to the `recharge_battery` function.

I think changing these functions to only use the discord_id instead of full Member objects is going to be more elegant in the long term. We only hand the functions what they need to do their job and thus they remain more modular. Is a bit of refactoring though.


ZebraNorth (commented):
Looks good!

HexLatex (approved):
Good

PR #462, opened 2024-06-25
Fix error when direct messaging the bot

ZebraNorth:
If the message is a DM then the author will have no `roles` property, so check for the existence of a guild prior to checking the roles.


3287 (approved):
3287 :: lgtm

PR #468, opened 2024-08-09
New logger; Refactor unit tests; DroneMember

ZebraNorth:


main.py
```diff
@@ -239,36 +239,39 @@ async def wrapper(message: discord.message):
...
-    LOGGER.info("Processing additional commands.")
-    await bot.process_commands(message)
+        # Replace the original message if it was altered by a listener.
+        if not isinstance(message.channel, discord.DMChannel):
+            await webhook.webhook_if_message_altered(message, message_copy)
+
+        # Report the error after the webhook so that the error message always follows the command,
+        # even if the original command message was deleted and replaced by the webhook.
+        if error:
+            await report_error(message.channel, str(error))
```
digitalAmy, line 272: I'm not sure if this will necessarily work. In certain situations, where the AI creates multiple messages back to back, they will appear in unexpected orders. I think this has to do with the asynchronicity (is that how you spell that?) of the whole API. Interactions with the discord API are not guaranteed to take effect in a specific order.

But no need to worry about that too much here. That is a general issue we may want to tackle separately later.

ZebraNorth: The issue here is that you should display the error message after running the webhook, otherwise the messages appear in the wrong order:

1) User sends command to be executed by a listener, eg a storage command: 1234 :: 0006 :: 1 :: Store the Hive Mxtress
2) The bot sends an error message: "You cannot store the Hive Mxtress"
3) The bot's identity enforcement listener runs.  This deletes message 1 and sends it again, with the default drone avatar and such.

At this point, the chat history looks like:

(Deleted message from User: 1234 :: 0006 :: 1 :: Store the Hive Mxtress)
Bot: "You cannot store the Hive Mxtress"
[Bot acting as user]: 1234 :: 0006 :: 1 :: Store the Hive Mxtress

The command is displayed after the error message, which is confusing.
The code above stores the error message instead of outputting it in step 2, and outputs it after step 3.  This solves the issue.

From looking at the pycord source, it looks like when sending a message it waits for a response from the server before continuing, so messages should appear in the order sent.

digitalAmy: oh you are probably right

test/mocks.py
```diff
@@ -0,0 +1,610 @@
...
+
+        The bot should usually only have one Cog, as added by the @cog() decorator.
+        '''
+
+        cogs = self.get_bot().cogs
+
+        if len(cogs) == 0:
+            raise Exception('Bot has no Cogs')
+
+        return cogs[next(iter(cogs))]
```
digitalAmy, line 105: I understand that this works, but is there a reason to use this instead of using `cogs[0]`? This may just be not knowing some arcane case where that would fail.

digitalAmy: No wait, why does this work? Doesn't `next()` already return the first element?

test/mocks.py
```diff
@@ -0,0 +1,610 @@
...
+
+        return record
+
+    def drone(self, drone_id: str | int, **kwargs) -> MagicMock:
+        '''
+        Create a mock drone record.
+        '''
+
+        if len(str(drone_id)) != 4:
+            raise RuntimeError('Drone IDs should be 4 digits.')
```
digitalAmy: This will lead to errors, if one enters a drone_id as an int like 0042, as the leading zeroes will be dropped. Is it maybe easier to just drop support for ints here, so we don't have to deal with those cases?

Otherwise I would kinda expect this function to perform padding with zeroes.

test/mocks.py
```diff
@@ -0,0 +1,610 @@
...
+        '''
+        Create a mock channel and add it to the guild.
+        '''
+
+        existing = get(self._guild.channels, name=name)
+
+        if existing:
+            return existing
+
+        # channel = MagicMock(spec=TextChannel)
```
digitalAmy: Should this still be here or is this a leftover from development?

HexLatex (approved):
yeah sure why not

PR #492, opened 2024-10-26
Consider name, nick, and global name when assigning an ID

### LINKED ISSUE #401: take global nickname into account when assigning an ID
When a user goes into assignment the AI does not look into their global nickname for a potential ID. We should change this but that may require an API upgrade.

ZebraNorth:
Closes #401


HexLatex (approved):

src/bot_utils.py
```diff
@@ -109,16 +110,25 @@ def predicate(ctx: Context) -> bool:
...
-    found = re.search(r"\d{4}", username)
-    if found is None:
-        return None
-    else:
-        return found.group()
+    # List the member's names in order of preference.
+    names = [member.nick, member.global_name, member.name]
+
+    # Filter out entries that are None.
+    names = [name for name in names if name is not None]
```
3287, line 122: This gets the job done, but definitely feels kinda clunky. Maybe consider this instead?
```py
names = list(filter(lambda x: x is not None, names))
```

ZebraNorth: tbh I find the lambda much more clunky

PR #496, opened 2024-10-27
Emoji IDs may be more than 18 digits

ZebraNorth:
Emojis have grown beyond 18 digits.  Make sure they don't get glitched.


3287:
closes #500

HexLatex (approved):

